PR #163, opened 2015-01-02
Changed Zoom Tool functionality to keep centred on the middle of the canvas instead of on the mouse

iirelu:
Playing around with this change for a bit felt a lot more convenient and consistent than the previous functionality, but I'm still making this a PR as it is a potentially significant user-facing change, so it probably needs feedback.

It definitely feels a lot better when using it as ctrl+middlemouse along with the rotating shift+middlemouse.

### DIFF
--- a/gui/viewmanip.py
+++ b/gui/viewmanip.py
@@ -79,8 +79,7 @@ class ZoomViewMode (gui.mode.OneshotDragMode):
             self.ACTION_NAME)
 
     def drag_update_cb(self, tdw, event, dx, dy):
-        tdw.scroll(-dx, -dy)
-        tdw.zoom(math.exp(dy/100.0), center=(event.x, event.y))
+        tdw.zoom(math.exp(dy/100.0), center=tdw.get_center())
         # TODO: Let modifiers constrain the zoom amount to
         #       the defined steps.
         self.doc.notify_view_changed()


0ion9:
Isn't this what it used to work like?
Personally, I'm concerned about maintaining the ability to quickly whiz out of one area of the image and into another. With the current zooming, it's very easy to go from point X to point Y: Shift+Mdrag (up until zoom level is satisfactory, then left/right until the correct part of image is shown) followed by another Shift+Mdrag(starting on the object of interest, and going down.).

After making the above change, this is not at all true: the process of first panning, then zooming, then finding what I wanted has slid offscreen and having to pan again, is error prone.

Have we tried the reverse behaviour? that is, leaving zooming as is and making rotation be around the cursor?

iirelu:
What you suggested is actually the first thing I tried. I came to the conclusion that continuous navigations (dragging the mouse around) and stepping navigations (scrollwheel and zoom in/out etc hotkeys) should logically follow two distinct rulesets:

 | Continuous | Stepping
--- | --- | ---
**Zoom** | Ctrl+middlemouse: Zooms into the **centre** | Scrollwheel: Zooms into the **cursor**
**Rotate** | Shift+middlemouse: Rotates around the **centre** | Shift+scrollwheel: Rotates around the **cursor**

This PR just changes the functionality of continuous zooming so Mypaint's navigations follow this table.

Thanks for your feedback, by the way. Out of curiosity, why can't you use the scrollwheel for the functionality you described as losing?

0ion9:
"scrollwheel"? What is this "scrollwheel" you speak of? ;) 
(my stylus doesn't have one, and using a mouse in my off hand is a bit problematic -- I want to use the keyboard way too much for it to be practical)

The table explains your thinking better, thanks for taking the time to put that together. I hadn't even considered that anyone would use a scrollwheel.

iirelu:
Actually, you're right. The old method *does* feel a lot better when using it with the tablet. I'll think about things some more and see what I come up with.

martinxyz:
> Have we tried the reverse behaviour? that is, leaving zooming as is and making rotation be around the cursor?

From memory, I think it started out with everything centered around the cursor. This is to support the "one hand tablet, other hand keyboard" workflow, especially for large (A4) tablets zoom at stylus is much more useful IMO. Then there was a user request to make rotation screen-centric, I think we got consensus that this is better (if you work in the upper left corner of an A4 tablet you may want to zoom there, but not to rotate half of the drawing out-of-screen).

achadwick:
I think we need change too... but I'm of completely the opposite view to @iirelu ... eek!
There's a lot to think about here. I'll try and explain my POV. For reference, here's the current behaviour. Surprise! it's all cursor-centric at the moment.

|  | A: Motion with modifier and button held | B: Scroll, with optional modifier | C: Accelerator key | D: Toolbar increase/decrease button
| --- | --- | --- | --- | ---
| **Zoom** | *Up/down:* a "flying zoom", where initial vertical position keeps current zoom[1]. *Left/right:* horizontal pan[1] | cursor | cursor[2] | cursor[3]
| **Rotate** | Rotation centred on initial cursor position | cursor[2] | cursor[2] | cursor[3]

And my opinions for what they're worth:

1. Weird behaviour, but kinda nice
2. Bad behaviour... some of the time. This feels mostly OK if the cursor is "fairly central" *AND* you're "using the pointer" in some way.
3. Basically always bad IMO. The user isn't pointing at the canvas here, but they *are* pointing and clicking. It's misleading to zoom at that pointer position.

----------------------------------------------------

So what would I like, in an ideal world? Like Maxy, I want to support a workflow where you're generally focussed on the cursor, with anything else sort of affecting things around where you're looking. However zooming and rotation can be invoked in so many ways now.

* **I'd like to leave leave column A unchanged please**. That's fine, for me.
* Column B and C could perhaps be cursor-centric *but only if* the last motion came from an in-proximity tablet device. Perhaps making it viewport-centric in other circumstances makes sense?
  * Good consequence: tablet scroll strips and wheels are usually on the pad, not the stylus. So the user can scroll when the pen is on their desk, behind their ear or whatever. Using proximity would allow us to be certain that the user is holding the pen.
  * Bad consequence: mice don't have proximity, but I *really* want Zoom A to be at the cursor
    * Solution, maybe?: "always be cursor-centric [for zooms] if it's a mouse scroll wheel"
  * PITA consideration: not all tablets have proximity.
* I *hate* the current behaviour in the D column. However a real problem here is that the implementation uses the stuff in column C, and there's no way of distinguishing them.
* If B and C can't both be made super-smart about devices, then keeping column B always cursor-centric and making C and D both always viewport-centric would be fine for me.

I think the "simple" approach in mine is exactly opposite to that @iirelu wants though :/

iirelu:
Discussion is good! It's not like this is more than 2 lines changed, anyway.

It looks like it may be a good idea to re-think this again later when there aren't more pressing issues at hand. I'm closing this PR for now, though.

Issue #145, opened 2014-11-26
Trim to Frame: only trims to whole-tile boundaries


achadwick:
Little papercut with the layer trimming feature as of 61ba5536ce66a5a98a79398c083017b47cc5f2d4: it only discards whole tiles. Steps to reproduce:

* Open the tool options panel (*Edit → Options for Current Mode*)
* Scribble something
* Click on the *Frame* tool in the toolbar
* Adjust the frame so that it intersects the scribble. For best reproduction of this bug, tweak all corners.
* Within the options panel, click the *Trim Layer to Frame* button.

Unless you're very lucky and have hit a 64x64px boundary in the adjustment phase, you'll see that only whole tiles are discarded:

![Trim to Frame results, showing bits of the layer still protruding beyond the frame](https://cloud.githubusercontent.com/assets/61299/5200427/e055fc32-755d-11e4-8e54-ab93ff9e926d.png)

MyPaint should zero pixels that lie outside the frame, for background tiles only.

PR #170, opened 2015-01-11
zero pixels outside of the frame when trimming

jy2wong:
A tentative fix for https://github.com/mypaint/mypaint/issues/145.

Tentative because I'm not sure what a "background tile" is or how that would affect the fixing of this issue.

I found another trim() function in lib/strokemap.py when I grepped for 'Only complete tiles'. Am I correct in thinking that that's a separate problem?

### DIFF
--- a/lib/layer.py
+++ b/lib/layer.py
@@ -3530,6 +3530,9 @@ class SurfaceBackedLayer (LayerBase):
         :type rect: tuple (x, y, w, h)
 
         Only complete tiles are discarded by this method.
+        If a tile is neither fully inside nor fully outside the
+        rectangle, the part of the tile outside the rectangle will be
+        cleared.
         """
         self._surface.trim(rect)
 
--- a/lib/tiledsurface.py
+++ b/lib/tiledsurface.py
@@ -164,6 +164,9 @@ class MyPaintSurface (object):
         :type rect: tuple (x, y, w, h)
 
         Only complete tiles are discarded by this method.
+        If a tile is neither fully inside nor fully outside the
+        rectangle, the part of the tile outside the rectangle will be
+        cleared.
         """
         x, y, w, h = rect
         logger.info("Trim %dx%d%+d%+d", w, h, x, y)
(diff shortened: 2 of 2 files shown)


achadwick:
This looks fairly good so far - there's just one problem I can see. If you do a sequence of trim operations and then undo them, there's some corruption visible:

170 trim corruption

I think you might be addressing tiles in the layer that are also on the undo stack. For the edges, there should be a copy/write/reinsert for tiles intersecting the edge which require pixel shaving. Have a look at other things in `lib/command.py` for similar things.

Don't worry about background tiles: the background is another layer entirely: it's conceptually infinite paper, so a trim on it should be a no-op. The method is just there in case we add something for guillotining all layers at once.

Yeah, the strokemap should be trimmed by this operation I guess. But it's a lot less visible, feel free to leave the edge "ragged" (i.e. just drop whole tiles) there.

(I'd be happy to take it from here, or to accept an updated PR)

jy2wong:
oof. Disregard. I looked through lib/command.py and didn't understand until I pushed the hacky thing above.

jy2wong:
Poked around the source code a bit. I can think of a couple ways to fix the undo problem, but I'm not comfortable making a decision like this in a codebase I'm barely familiar with (and I'm not convinced that these are the most reasonable solutions, either).

1. have MyPaintSurface's implementation of save_snapshot() make a deep copy of the tiledict (this is a 2 liner and I've confirmed that it works, but does this result in an unacceptable increase in memory usage?)
2. expose actual tiles in lib/layer.py (with something analogous to get_tile_coords()?) and keep track of something like self.old_edge_tiles in lib/command.py's TrimLayer class (seems reasonable to me, but perhaps there's some sensible reason that get_tiles() doesn't already exist?)

achadwick:
I forget bits of the codebase myself :fearful: In this case, the tile access context manager. Doing

```
    with layer.tile_request(tx, ty, readonly=False) as rgba:
        rgba[1, 2, :] = 0
```

and the existing `save_snapshot()` will get you the copy-on-write semantics that undo and redo want. Tile requests like that work best if they're only one per tile.

Sorry about the fiddliness here. This is fairly deep-end stuff now.

achadwick:
If you fix up the two things above and wrap up the changes in a fresh branch or a squashed (force-pushed?) version of this one, I'd be happy to accept these changes. Thanks for working out the ranges and testing those out!

(I'm not very familiar with the request side of a pull request, so I'm not sure how to revise them for neatness. Forgive me if I'm saying anything stupid.)

jy2wong:
Modified, cleaned up a bit, rebased and force pushed. This pull request page looks a bit odd now, though.

jy2wong:
Sorry to bother, but it's been a few days now. Bumping this just in case it slipped through the cracks. If you're busy with other stuff I totally understand. Take your time! I just figured this shouldn't get lost.

achadwick:
Thanks for the reminder, been a little busy trying to make MyPaint run properly on Windows.

The updated PR looks great now, and undo/redo works well. Thanks for the contribution!

Issue #65, opened 2014-07-29
Logging: `Found old-style configuration` message needs to make it clearer that the old config is being used


achadwick:
Breaking out part of the RFH in #64: the logging messages regarding old-style `.mypaint` config areas needs to be improved. We currently say

```
INFO: mypaint: Found old-style configuration in u'/home/andrewc/.mypaint/'
INFO: mypaint: This can be migrated to $XDG_CONFIG_HOME and $XDG_DATA_HOME if you wish.
INFO: mypaint: See the XDG Base Directory Specification for info.
```

This doesn't explain that the old config area is still being used (and that this is _absolutely fine, no need to take any action right now_).

Maybe we don't need to say anything at all, or just deprioritize these messages to DEBUG so they won't appear by default.

achadwick:
Closed by PR.

PR #176, opened 2015-01-21
make it clear when the old configuration area is being used

jy2wong:
Addresses issue #65 and also tweaks wording a bit to better reflect what conforming to the XDG basedir spec means.

I personally am okay with such a message being info, because I like putting my stuff in $XDG_{CONFIG,DATA}_HOME whenever possible and in theory this message would have told me that I could move mypaint configs into those directories (in practice it got buried in GTK warning messages).

This is outside the scope of this issue, but it also wouldn't hurt to add something to a README about what needs to move where if you want to migrate to $XDG_{CONFIG,DATA}_HOME.

### DIFF
--- a/mypaint.py
+++ b/mypaint.py
@@ -196,9 +196,10 @@ def get_paths():
         if not os.path.isdir(old_confpath):
             old_confpath = None
         else:
-            logger.info("Found old-style configuration in %r", old_confpath)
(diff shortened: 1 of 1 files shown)


achadwick:
Thanks for the updated strings. Much clearer now!

Issue #146, opened 2014-11-27
<Primary> should change to <ctrl> or <command> depending on the current OS


iirelu:
\<Primary> is a bit unfriendly as something user-facing. It's good for underlying code, as it makes things easily interoperable between OSes, but the user should see the more familiar term for their platform.

PR #180, opened 2015-01-22
Use localized strings for key bindings in prefs dialog

jy2wong:
This should fix issue #146.

Tested saving preferences, adding new shortcuts, removing existing shortcuts and the shortcut cheatsheet thing that appears in the bottom right corner of the window ("With Ctrl held down...").

Spun up mypaint with a french locale as well. Modifier keys (ctrl, alt, shift) are being translated by gtk.accelerator_get_label(). Button/Btn will need support in po/, but if strings like "Button1" already have a translation, then it should be pretty easy to do even if you don't speak the language in question.

### DIFF
--- a/gui/buttonmap.py
+++ b/gui/buttonmap.py
@@ -40,7 +40,7 @@ def button_press_name(button, mods):
     return modif_name + "Button%d" % (button,)
 
 
-def button_press_displayname(button, mods):
+def button_press_displayname(button, mods, shorten = False):
     """Converts a button number & modifier mask to a localized unicode string.
     """
     button = int(button)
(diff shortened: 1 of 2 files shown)


iirelu:
Ooh, does look like an interesting change!
* Before: blender
* After: blender

I'd love for the UI to just make assumptions about what's left mouse, right mouse, and middle mouse instead of the vague "ButtonX" it has now, but that can be changed later.

I'll wait to see if @achadwick has anything to say about this, but the PR looks good to me!

achadwick:
@iirelu Left/right/middle only really applies to mice. We have to say "ButtonN" *and* use digits for devices with more then three buttons anyway.

achadwick:
@jy2wong 
Is there any risk that "mouse button" (or "device button") may have a different translation to any other uses we have of _("Button") in some languages (thinking "onscreen button" senses here)?
I may be being paranoid.

* If so, I'd prefer https://github.com/jy2wong/mypaint/commit/890ee2f97dc14934e4c1a9b6b1be6ac3efc5944e#diff-7bf3f877e0bfc3b6f9791f03a5158362L56 to use a different long template for the abbreviated and non-abbreviated cases.
* If not, the `#TRANSLATORS:` comment should be removed here because the following line does not call any gettext functions.

I'm leaning towards the 2nd interpretation, but horribly aware that I speak too few languages to really know.

Other than that, the rest of the PR looks fine. If you can fix up the patch appropriately, I'be be happy to merge it.

jy2wong:
Quick survey suggests both senses of "button" map to the same word in chinese, finnish, polish, somali, italian, french, arabic and japanese.

I've moved the #TRANSLATORS: comment up to the line it really should be on and made it even more clear for future translators that "button" is used in the mouse sense juuuuuust in case.

achadwick:
@jy2wong 
Thanks so much for the update PR, and especially for doing the legwork with all those languages: it's greatly appreciated! I wouldn't know where to start.

Issue #129, opened 2014-11-12
Color wrong when opening PNGs with gAMA chunk


ydahhrk:
This is supposed to be a valid PNG (according to Gimp):

![Valid PNG](https://dl.dropboxusercontent.com/u/95836775/MyPaint/000.png)

I painted the central blob with color #BBDDAA.

If you open the image with Gimp or Krita, pick the green color and have a look at the hex value, you will see "#BBDDAA". I also tested it on a couple of closed source drawing programs, and they also report "#BBDDAA".

If you do the same with MyPaint, you get color "#BCDEAB" in the "Color Details" window.

There seems to be a problem with MyPaint's PNG loader. This also affects ORA images since they use PNG internally.

Not all PNG images I've seen suffer from this problem:

- PNG images saved by MyPaint are loaded by MyPaint correctly.
- ORA images saved by MyPaint are loaded by MyPaint correctly.
- PNG images saved by Gimp are loaded by MyPaint correctly.
- ORA images saved by Gimp are loaded by MyPaint **incorrectly**.

(That sample image was outputted by Gimp's ORA exporter; I unzipped the ORA to get the PNG).

The corruption is accumulative: MyPaint will save color #BCDEAB, since that is what it sees. Gimp will then save #BCDEAB, but MyPaint will again load it as something else (such as #BDDFAC). After several saves, the corruption is very noticeable:

![Start](https://dl.dropboxusercontent.com/u/95836775/MyPaint/start.png)

![End](https://dl.dropboxusercontent.com/u/95836775/MyPaint/end.png)

I first saw this on Ubuntu 12.04's MyPaint version. I just tested it in the master Github version and it's still a problem.

D:

jy2wong:
I've verified the default gAMA and cHRM values against http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.sRGB . It's not clear from the standard whether this is exactly equivalent to using sRGB, or just values that are close enough. It's also not clear to me whether these are sensible defaults, or whether it's sensible to have defaults at all.

As far as I can tell, there's nothing in the specification about what you're supposed to do with just a gAMA chunk. It just says that "the absence of a cHRM chunk indicates that the image's primary colors are device-dependent."

If you comment out the following block and then remove the word else from the next 3-line block (that is, you completely ignore the gAMA chunk and default cHRM values and pretend the file uses sRGB), the colour does come out as #BBDDAA. Interesting.
[...]

PR #183, opened 2015-01-28
when reading PNG files, ignore gAMA and cHRM chunks in a special case

jy2wong:
for issue #129.

The 'sRGB' gAMA and cHRM values will never result in exactly the same colours as sRGB, since sRGB uses a more complicated gamma curve that can't be represented by a single number.

What do you think? Is it reasonable to assume sRGB if the read gAMA and cHRM chunks suggest the default sRGB approximation values?

### DIFF
--- a/lib/fastpng.hpp
+++ b/lib/fastpng.hpp
@@ -5,6 +5,8 @@
 #include "lcms2.h"
 #endif
 
+#include <cmath>
+
 #ifndef SWIG
 static void png_write_error_callback(png_structp png_save_ptr, png_const_charp error_msg)
 {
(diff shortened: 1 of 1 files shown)


achadwick:
I sort of like this approach in a way, but the sRGB assumptions and gamma curve should probably only be applied when we have a real `sRGB` chunk in there.

Maybe either a `gAMA` chunk or a `cHRM` chunk without the other should be considered meaningless for doing generic RGB conversions: I'm thinking we should do nothing in this case, i.e. attempt no conversions. If GIMP chooses to write half of the necessary colorspace information, that's _its_ problem and not ours!

jy2wong:
Makes sense, and seems way less hacky than my solution. Going by the discussion in the original issue, it looks like you're already planning to do some code cleaning in the general area, so if you want to take care of this I'd be happy to close this pull request (or to keep it open and fix things properly).

achadwick:
In the end, I've sort of adapted 

If we assume that an explicitly tagged sRGB file's pixel data can be used verbatim, we can assume that gamma=2.2 means roughly the same thing. Just as GIMP does when it's writing files from its default uncalibrated internal representation. The two tone curves aren't really identical, but are quite similar.

(Close enough for <s>government work</s> our purposes, I guess.)

PR #186, opened 2015-01-29
Version Call in the Crash Log Change

odysseywestra:
The version number in the crash log now includes the build date, and
the git version. This should make it easier to find out which git
versions are affecty by the reported issue.


iirelu:
Oh, thanks! I briefly tried to find some way to get the full git version, but I didn't realise the application singleton had its own version definition. Will test and merge.

iirelu:
`Mypaint version: 1.1.1-alpha.20150129+git.52bf64d`

Looks alright to me! If it needs tweaking, that can be done in application itself.

PR #187, opened 2015-02-01
Edit Composting Mode Tool Tips.

odysseywestra:
I've edited the tooltips strings in ```/lib/layer.py``` for the Composting Mode 
descriptions. These explanations are a bit more indepth to help artists understand
what they do. Plus the tooltip for the Destination compositing layer didn't specify that 
they can be used for masking out or in object.

TODO: Probably search for a way to insert text wrapping to the tooltips instead of 
manually typing the newlines in the strings.

### DIFF
--- a/lib/layer.py
+++ b/lib/layer.py
@@ -79,85 +79,114 @@ MODE_STRINGS = {
     # Group modes
     PASS_THROUGH_MODE: (
         _("Pass-through"),
-        _("Group contents apply directly to the group's backdrop"),
+        _("Group Layer's compositing applys directly to the \n"
+          "layer(s) below."),
         ),
     # Standard blend modes (using src-over compositing)
     mypaintlib.CombineNormal: (
         _("Normal"),
-        _("The top layer only, without blending colors.")),
+        _("No compositing applied to the layer.")),
     mypaintlib.CombineMultiply: (
         _("Multiply"),
-        _("Similar to loading two slides into a projector and "
-          "projecting the combined result.")),
+        _("Darkens by multiplying the luminance value of \n"
+          "the top layer and the layer below. Similar to \n"
+          "stacking two slides, and projecting the combined \n"
+          "result. This is the inverse of 'Screen'.")),
     mypaintlib.CombineScreen: (
         _("Screen"),
-        _("Like shining two separate slide projectors onto a screen "
-          "simultaneously. This is the inverse of 'Multiply'.")),
+        _("Lighens by multiplying the inverse value of the \n"
+          "colors of the top layer and layer below. Similar \n"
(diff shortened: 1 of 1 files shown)


achadwick:
Hi - I think we need to work on this a bit more. We also need to keep it terse so that the tooltip can be read quickly.

BTW, please don't insert `\n` into the tooltips. We don't know how wide the windowing system will make them, so the resultant string may format really weirdly.

If we need more gloss than what we have for the operations, it's possible to use a diagram+title+description tooltip with not much extra coding: similar to what we do for the current tool indicator I think. The layout can look like:

```
DIAGRAM   TITLE.TITLE.TITLE     ← in bold
DIAGRAM
DIAGRAM   DESCRIPTION.DESCRIPTION.DESCRIPTION.DESCR
DIAGRAM   IPTION.DESCRIPTION.DESCRIPTION
```

The image/diagram is approx 48x48 px, and it' optional.

achadwick:
(Sorry about all the line comments appearing in the thread - I think they vanish from the PR when you make a force-push to the branch)

iirelu:
Hm. It definitely looks better with newlines than without, but it's a suboptimal solution. Preferably the tooltip drawing code should manage the maximum width.

Without newlines (this is the old text) things can get *really* long:
blender

A thinner tooltip looks a lot better:
b

achadwick:
Oh wow the unwrapped one looks bad. Clearly my memory is at fault: I thought themes and GTK were fairly sensible about long tooltips.

Arguably long tooltips are problematic in and of themself.
Maybe the job of explaining the modes in detail could be better done with a ? button to the right of the combo and a click-or-hover popover. Hover for mice, click for touch or tablet...

iirelu:
I think paragraph-length tooltips are fine. Tooltips are a near-universal way for the user to get more information about a specific thing and arguably getting rid of them is a lot worse than just having long ones.

A helper function to format tooltip strings to a max width might work, unless GTK has something for doing that itself.

achadwick:
Okay, let's keep a tooltip for the combo. But strongly favour headline and _short_ para.

FWIW, Adwaita does a decent job on my system; I guess it's quite theme-dependent.

tt

We might be able to do something cunning in a `query-tooltip` handler and `gtk_tooltip_set_custom()`, like set a label's maximum width. But I really don't want to put in hacky newlines that might screw up in themes with narrow markup-flavour tooltips.

iirelu:
Hm, if themes can change the tooltip's maximum width surely we can too? [A quick google suggests yes](http://stackoverflow.com/questions/25701036/force-label-line-wrap).

odysseywestra:
I'm going to close this pull requrest, and open up an issue ticket cause this will require more work.

Issue #165, opened 2015-01-05
MyPaint Not Saving or Exporting PNG Files Correctly


odysseywestra:
I've notice this when I was working my last drawing. However, it seems to only effect when I just save it as PNG transparent(.png) and not when I save it as a Multiple PNG transparent(.XXX.png), or as an .ora. I even went back a few commits, and it seems to not change. Is anybody else having a problem with this as well?

System information: Linux-3.16.0-28-generic-x86_64-with-Ubuntu-14.10-utopic
Version: 1.1.1-alpha.20150103+git.6947ad9

iirelu:
Huh, you're right! I guess that's been broken for a while and just nobody's noticed.

jy2wong:
The first three images were done with 'Show background' unchecked in the layers panel and in a fresh instance with a fresh config. Files were saved by typing in the filename with no extension and selecting "PNG transparent (*.png)" in the dropdown.

First export:
![a](https://cloud.githubusercontent.com/assets/4390573/5708816/4f6b9ebe-9a62-11e4-8d37-e00bef24ca70.png)
First save as:
![b](https://cloud.githubusercontent.com/assets/4390573/5708817/52563166-9a62-11e4-8038-fe5c5d93cc84.png)
First save:
![c_orig](https://cloud.githubusercontent.com/assets/4390573/5708995/eed0652e-9a63-11e4-9272-5f432679a52c.png)

Huh. Seems like saving with transparency works the first time you try to save (by any means). I modified the third image, tried saving it immediately afterwards and got
![c_second](https://cloud.githubusercontent.com/assets/4390573/5709004/f9868480-9a63-11e4-952d-34bdc4349c3f.png)
It appears that all the transparent pixels were turned black.

The same thing seems to happen if you save something as a .ora first, then try to save as a transparent PNG.

odysseywestra:
So I tried it again and it seems thats the ```[ ]show background```option in the layers dialog is having an effect on saving transparency.

When the ```[x] show background``` is check and I tried to export, or save it as a transparent .png, it saves it as a png with a background. When the ```[ ] show background``` is unchecked, and I tried to export, or save it as a transparent, it saves as a transparent png.

Another Odd thing I notice is when I just save or save as it as a png with background, it fills it with black. It even worse since save, just repeat what 

I renamed the bug as to better reflect it.

So far the Open Raster File is not effected by this bug.

jy2wong:
[...]

PR #188, opened 2015-02-03
Correctly handle PNG transparency when saving

jy2wong:
For issue #165. The parameter that governed whether or not blit_tile_into() kept the background in or not was inconsistently named (it was 'background' in half the source code, and 'render_background' in the other half). I suspect the regression made its way in because it's...nontrivial to keep track of all the parameters weaving in and out of kwargs/options on the way from gui/filehandling.py to lib/fastpng.hpp.

### DIFF
--- a/gui/filehandling.py
+++ b/gui/filehandling.py
@@ -333,7 +333,7 @@ class FileHandler(object):
             recent_data.mime_type = mime_type
             recent_mgr.add_full(uri, recent_data)
         if not thumbnail_pixbuf:
-            options["background"] = not options.get("alpha", False)
+            options["render_background"] = not options.get("alpha", False)
             thumbnail_pixbuf = self.doc.model.render_thumbnail(**options)
         helpers.freedesktop_thumbnail(filename, thumbnail_pixbuf)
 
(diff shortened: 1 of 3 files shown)


achadwick:
The `SAVE_FORMAT_PNGAUTO` has a good, simple "Do What I See" nature. I like it.

Thanks for doing the detective work re. the parameters. Yeah, that API and parameter chain is an old horror that needs sorting out and/or documenting. One day.

Let me know if you want commit access, by the way. I can add you to @mypaint/collaborators.

jy2wong:
Oops, I never noticed because the items showed up in the same order as before on my machine. That's certainly not something to rely on, though. And thanks for pointing me towards OrderedDict; I didn't know it was a thing. Anyway, this is fixed up now.

As for commit access, I wouldn't mind it, but I'd probably end up pull-requesting anything greater than 2 lines and/or possibly debatable anyway because I think it's generally good to have more than one set of eyeballs looking at incoming code.

achadwick:
Invite sent. I'm happy to review things via PR always, but you may find team membership useful for just pushing oneliner fixes. I think (but I'm not sure) that you'll get the ability to merge PRs too.

0ion9:
@achadwick : Answer seems to be yes -- see https://help.github.com/articles/permission-levels-for-a-user-account-repository/ , various operations on PRs ,including merging, are on the list.

PR #196, opened 2015-02-11
fix for ora files without a default selected layer

vurentjie:
i picked up an issue if i save ora files in krita, there is no default selected layer,
then when re-opening in mypaint and trying to select the layer it breaks, 

  File "/usr/local/share/mypaint/gui/layerswindow.py", line 469, in _view_button_press_cb
    docmodel.select_layer(path=click_layerpath)
  File "/usr/local/share/mypaint/lib/document.py", line 380, in select_layer
    self.do(command.SelectLayer(self, path=sel_path))
  File "/usr/local/share/mypaint/lib/command.py", line 715, in __init__
    self.prev_path = layers.canonpath(path=layers.get_current_path())
  File "/usr/local/share/mypaint/lib/layer.py", line 2642, in canonpath
    (path,))
ValueError: path=() is root: must be descendent

### DIFF
--- a/lib/command.py
+++ b/lib/command.py
@@ -710,7 +710,7 @@ class SelectLayer (Command):
         super(SelectLayer, self).__init__(doc, **kwds)
         layers = self.doc.layer_stack
         self.path = layers.canonpath(index=index, path=path, layer=layer)
-        self.prev_path = layers.canonpath(path=layers.get_current_path())
+        self.prev_path = layers.canonpath(path=layers.get_current_path(),usecurrent=True)
 
     def redo(self):
         layers = self.doc.layer_stack
--- a/lib/layer.py
+++ b/lib/layer.py
@@ -2732,7 +2732,7 @@ class RootLayerStack (LayerStack):
         """
         if path is not None:
             layer = self.deepget(path)
-            if layer is self:
+            if layer is self and not usecurrent:
                 raise ValueError("path=%r is root: must be descendent" %
                                  (path,))
             if layer is not None:


achadwick:
I can't reproduce this. Could you paste the stack.xml out of an affected Krita .ora file between triple-backticks please?

    ```
    <xml-here/>
    ```

I want to see how it's marked up and see if I can build myself a test case.

(OpenRaster (.ora) files are glorified Zip archives)

achadwick:
I'm not sure we should address this in the undo stack entry, since MyPaint needs a layer always selected, i.e. current for painting. If necessary, we should force it upon loading the file non-undoably - and I thought that's what the code does atm.

vurentjie:
to force when parsing the xml would probably be better actually

this is what krita generated, 

```
<image w="4992" version="0.0.1" h="2112" yres="75" xres="75">
 <stack opacity="1" visibility="visible" name="Group" composite-op="svg:src-over">
  <layer opacity="1" visibility="visible" name="mask" composite-op="svg:screen" src="data/layer4.png"/>
  <stack opacity="1" visibility="visible" name="ink " composite-op="svg:src-over">
   <layer opacity="1" visibility="visible" name="ink" composite-op="svg:src-over" src="data/layer3.png"/>
  </stack>
  <stack opacity="1" visibility="visible" name="colour " composite-op="svg:src-over">
   <layer opacity="1" visibility="visible" name="colour" composite-op="svg:src-over" src="data/layer2.png"/>
  </stack>
  <stack opacity="1" visibility="visible" name="sketch " composite-op="svg:src-over">
   <layer opacity="1" visibility="visible" name="sketch" composite-op="svg:src-over" src="data/layer1.png"/>
  </stack>
  <layer opacity="1" visibility="visible" name="background" composite-op="svg:src-over" src="data/layer0.png"/>
 </stack>
</image>
```

vurentjie:
here is a ora file that has this problem

http://www.vuren.net/2015-02-11_6.ora

vurentjie:
well i had a look at the load_from_openraster method in layers.py

i fixed it there by changing the intial loop check to this, which just selects the top level layer:

```python
for path, loaded_layer in self.deepenumerate():
    if loaded_layer.initially_selected:
        selected_path = path
    elif selected_path is None:
      selected_path = path
    num_loaded += 1
```

but i don't really understand why the final check is not working, which is probably where the problem lies:

```python
if not selected_path:
    selected_path = [max(0, num_layers-1)]
```

i'll have to leave this to you

achadwick:
Reproduced with the test case image, thanks for making it.

achadwick:
Looks like some of that code came from the days when MyPaint didn't have nested layers. Thanks for the investigation and sample code, @vurentjie :+1:

PR #207, opened 2015-02-16
Viewfinder style change

jy2wong:
new_viewfinder

I was a bit concerned about what the cairo path juggling in the new `draw_draggable_path_drop_shadow()` would mean for performance, though it doesn't seem noticeably slower on my machine.

The extra `math.ceil` calls are there because there are occasional rendering artifacts without them. I'm not really sure what they actually fix, since I thought that rounding protection was what the `+ 4` was for.

I also wanted to have a chat about when/if the viewfinder frame should change colour. I think it makes sense to have it turn the active colour (yellow) when it's hovered over and while it's being dragged. But what about when a user is zooming/rotating through the canvas?

### DIFF
--- a/gui/drawutils.py
+++ b/gui/drawutils.py
@@ -431,11 +431,18 @@ def render_round_floating_color_chip(cr, x, y, color, radius):
 
 
 def draw_draggable_edge_drop_shadow(cr, p0, p1, width):
-    """Draw the drop shadown for an edge which can be dragged"""
-
-    cr.save()
+    """Draw the drop shadow for an edge which can be dragged, then clear the path"""
     x0, y0 = p0
     x1, y1 = p1
+    cr.move_to(x0, y0)
+    cr.line_to(x1, y1)
+    draw_draggable_path_drop_shadow(cr, width)
+    cr.new_path()
+
+def draw_draggable_path_drop_shadow(cr, width):
+    """Draw the drop shadow for all draggable edges in the current path, without clearing the current path"""
+
+    cr.save()
 
     # Drop shadow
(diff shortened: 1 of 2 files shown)


achadwick:
Hmm. I wonder if the lines should be thicker now; against a light background like the default cream/white it's a little invisible on a highish-res screen :/

I think the viewfinder is best though of as something the user can pick up and grab rather than something that flashes in response to external zooming or rotating activity: so I guess it's best to prelight it when hovering or dragging the cursor over it, as if it were a frame edge or a button?

iirelu:
IMHO it looks fine right now. Making everything grey just because it fits in better is a pain for readability.

jy2wong:
Another possibility is to change the style to match all the other draggable things, but keep it green.
green

iirelu:
That sounds like a good idea.

achadwick:
Seems good. In a case like this I think we need an accent colour. By the way, the panel is somewhat inspired by Paint Tool SAI's preview windows, and they use red for theirs in the screenshots I've seen.

We could reasonably use the orange-yellow or a redder but still light colour based on it.

iirelu:
Well, we can't really have our own colour scheme due to the inherent nature of what our program does. It's best to be *mostly* grey and leave colour down to the user.

achadwick:
The way of doing that would be to use the user's theme's accent colour, perhaps with an HCY blend. I kinda like the effect of just increasing the weight, though.

2px viewfinder

iirelu:
Hrrm, I still preferred the old green. Grey just doesn't stand out, no matter what you do.

PR #208, opened 2015-02-16
Use native cache directories for thumbnails in Linux and Mac OS X

jy2wong:
I don't usually have access to any OSX machines, so it'd be cool if someone with a Mac tested this out. Naming the directory 'MyPaint' seems to blend in better, but may cause headaches down the road. I can change that to 'mypaint' if it seems like it would cause too much trouble.

As for the Linux-y changes...how many applications actually use `~/.thumbnails` these days? Does anybody actually prefer using it to following the thumbnail spec? Is it better to just create `$XDG_CACHE_HOME/thumbnails` or `~/.cache/thumbnails` if needed?

### DIFF
--- a/lib/helpers.py
+++ b/lib/helpers.py
@@ -148,7 +148,7 @@ def gdkpixbuf2numpy(pixbuf):
 
 
 def freedesktop_thumbnail(filename, pixbuf=None):
-    """Fetch or (re-)generate the thumbnail in ~/.thumbnails.
+    """Fetch or (re-)generate the thumbnail in $XDG_CACHE_HOME/thumbnails.
 
     If there is no thumbnail for the specified filename, a new
     thumbnail will be generated and stored according to the FDO spec.
@@ -167,11 +167,8 @@ def freedesktop_thumbnail(filename, pixbuf=None):
     logger.debug("thumb: uri=%r", uri)
     file_hash = hashlib.md5(uri).hexdigest()
 
-    if sys.platform == 'win32':
-        base_directory = os.path.join(GLib.get_user_data_dir().decode('utf-8'),
-                                      'mypaint', 'thumbnails')
-    else:
-        base_directory = expanduser_unicode(u'~/.thumbnails')
+    base_directory = os.path.join(GLib.get_user_cache_dir().decode('utf-8'),
+                                  'thumbnails')
 
     directory = os.path.join(base_directory, 'normal')
     tb_filename_normal = os.path.join(directory, file_hash) + '.png'


achadwick:
Yes, let's fix it up for Linux at least. `~/.thumbnails` must come from some ancient spec.

I think (though I defer to @tumagonx here) that the Windows side of things should use `$XDG_CACHE_HOME/thumbnails` too. It'd be a fairly sensible place after all:

```
testuser@win7test MINGW32 ~
$ python2
Python 2.7.9 (default, Dec 26 2014, 00:24:06)
[GCC 4.9.2] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from gi.repository import GLib
>>> GLib.get_user_cache_dir()
'C:\\Users\\testuser\\AppData\\Local\\Microsoft\\Windows\\Temporary Internet Files'
>>> GLib.get_user_data_dir()
'C:\\Users\\testuser\\AppData\\Local'
>>>
```

a.k.a. CSIDL_INTERNET_CACHE which seems OK for little throwaway cache files like these. It'd be compatible with other FreeDesktop.org apps doing the same thing too, which might be helpful.

achadwick:
Regarding OSX: I kinda want to see where `GLib.get_user_cache_dir()` goes on a regular OSX/darwin box (with Apple-blessed Python) first. I'm guessing @jy2wong would be using it if it seemed sensible, but I'm not sure what the calls actually give you on OS X - I want to be sure the default is unhelpful first.

I'm not an OS X user at all, I must ask. Paging @bunk3m, @Pinzgauer, @ryandesign and any other regular/occasional OSX users: please can you tell us what `GLib.get_user_cache_dir()` and `GLib.get_user_data_dir()` return on your computer by running Python as above? It'll help us make a decision.

(IMO we should strive to use the same code path on all systems generally. Also, GLib is supposed to spit out _what you should use_ for the system it's running on and we should just trust it. No ifs, no buts, and no platform tests.)

("FreeDesktop" is an organization defining a common set of standards for Open Source apps to adhere to.)

bunk3m:
Hey @achadwick here is what I get in OSX 10.9.5.

    $ python -V
       Python 2.7.5

    $ python
    Python 2.7.5 (default, Mar  9 2014, 22:15:05) 
    [GCC 4.2.1 Compatible Apple LLVM 5.0 (clang-500.0.68)] on darwin
    Type "help", "copyright", "credits" or "license" for more information.
    
    >>> GLib.get_user_cache_dir()
    File "<stdin>", line 1, in <module>
    NameError: name 'GLib' is not defined

and then

    >>> GLib.get_user_data_dir()
    File "<stdin>", line 1, in <module>
    NameError: name 'GLib' is not defined

Hope this helps.  B.

jy2wong:
@bunk3m, try
```python
from gi.repository import GLib
GLib.get_user_cache_dir()
GLib.get_user_data_dir()
```

tedil:
You probably need to import glib first.
try:
`import glib`
`glib.get_user_cache_dir()`
`glib.get_user_data_dir()`

(or maybe 'Glib' or 'GLib' etc instead of 'glib')

bunk3m:
Sorry everyone but I haven't got a clue what you mean I should be doing.
I don't code but I know my way around the terminal a bit and I'm not afraid to try.  I downloaded GitHub gui, cloned mypaint, tried to run scon but it didn't work.  Now I'm completely confused and stuck.

@jy2wong where would I run that?

@tedil where would I import glib?  Is this from within Python?

bunk3m:
OK I think I might have figured out what you wanted me to do.  Unfortunately it didn't work.
> $ python
Python 2.7.5 (default, Mar  9 2014, 22:15:05) 
[GCC 4.2.1 Compatible Apple LLVM 5.0 (clang-500.0.68)] on darwin
Type "help", "copyright", "credits" or "license" for more information.

> from gi.repository import GLib
  File "<stdin>", line 1, in <module>
ImportError: No module named gi.repository

> from gi.repository import glib
  File "<stdin>", line 1, in <module>
ImportError: No module named gi.repository

> import glib
  File "<stdin>", line 1, in <module>
ImportError: No module named glib

achadwick:
@bunk3m Looks like you're missing GTK. Please follow the instructions at https://github.com/mypaint/mypaint/blob/master/README_OSX.md where it says:

> Install dependencies:
> 
>     sudo port install gtk3
>     sudo port install json-c
>     sudo port install py27-numpy
>     sudo port install py27-scipy
>     sudo port install lcms2
>     sudo port install py27-gobject3
>     sudo port install hicolor-icon-theme

You will need MacPorts for this: https://en.wikipedia.org/wiki/MacPorts and https://www.macports.org/

bunk3m:
Thanks again @achadwick 
I thought I'd followed all the instructions on that page.  But it appears that I didn't have gtk3.  It is installed now and then I followed the remaining commands but it died at ... 

> $ git submodule update --init --force
fatal: repository '../libmypaint.git' does not exist
Clone of '../libmypaint.git' into submodule path 'brushlib' failed

Did I miss this somewhere in the instructions?

iirelu:
@bunk3m Could you report a new issue or do this elsewhere? You've spammed an issue and a PR with this already, and it's completely unrelated.

bunk3m:
My apologies @iirelu . I didn't my helping the project was spam.

As my help isn't appropriate, I'll just stop.  

Sorry and bye.

iirelu:
@bunk3m No, I'm sorry, I just meant it would be better suited in maybe the #mypaint IRC channel or a new issue.

jy2wong:
Okay, I have an answer!

```python
from gi.repository import GLib
GLib.get_user_cache_dir()
GLib.get_user_data_dir()
```

on a Mac with https://trac.macports.org/browser/trunk/dports/python/py-gobject3/Portfile installed prints out

```
'/Users/jsmith/.cache'
'/Users/jsmith/.local/share'
```

Also,
```bash
$ echo $XDG_CONFIG_HOME $XDG_DATA_HOME $XDG_CACHE_HOME

```

On the machine of the person I asked, there was fontconfig stuff in `~/.cache` and `~/.local/share` existed (although it was empty).

I initially went with `~/Library/Caches` because the docs suggested that the cache dir would be `~/.cache` and I wasn't sure if that directory was used at all. I still like the idea of using the ~Mac native~ location of `~/Library/Caches` (even Firefox does, according to https://support.mozilla.org/en-US/questions/983090), but since `~/.cache` is already being used, maybe that's a reasonable spot too.

jy2wong:
Decided to do away with `~/.thumbnails` altogether, and have Windows and all the unices use `$XDG_CACHE_HOME/thumbnails`.

achadwick:
Thanks for the investigation @jy2wong and for the revised patch. I'm definitely in favour of getting rid of ~/.thumbnails. There's probably enough free software out there using it to make $XDG_CACHE_HOME fairly unsurprising for OSX users, I guess.

If it's an issue for users that default-config fd.o software writes to ~/.cache, then it should either be fixed by Apple (set the XDG vars in the user's environment), or by GLib (make g_get_user_cache_dir() on OSX return Apple's preferred locations, as it does Windows and MS's - although in the latter case you get the unlovely `%LOCALAPPDATA%\Microsoft\Windows\Temporary Internet Files\` for user cache, as I've discovered for #264 :unamused::confounded:).

Given that MyPaint itself may be using the g_get_user_cache_dir() cache area in the near future, I think this is the right approach to thumbnail storage too.

Issue #87, opened 2014-08-21
Movement tools deselect themselves after a single use


achadwick:
@iirelu reports in #81 that:
> [...] as it stands the movement tools are already entirely useless thanks to the fact that they deselect themselves after a single use, rendering them entirely pointless as tools.

I don't agree that they're entirely pointless or entirely useless, but let's fix the self-deselection. On reflection, that's very unhelpful behaviour.

PR #215, opened 2015-02-24
Do not deselect movement tools after one use

jy2wong:
Where "movement tools" means Move Layer, Pan View, Zoom View or Rotate
View modes. Addresses mypaint/mypaint#87.

Allow the popup menu to be brought up while in the above three modes

Fix a small typo in a comment in gui/mode.py

### DIFF
--- a/gui/colorpicker.py
+++ b/gui/colorpicker.py
@@ -74,7 +74,7 @@ class ColorPickMode (gui.mode.OneshotDragMode):
         return _(u"Set the color used for painting")
 
     def __init__(self, **kwds):
-        super(ColorPickMode, self).__init__(**kwds)
+        super(ColorPickMode, self).__init__(unmodified_persist=False, **kwds)
         self._overlay = None
         self._started_from_key_press = self.ignore_modifiers
         self._start_drag_on_next_motion_event = False
--- a/gui/document.py
+++ b/gui/document.py
@@ -663,7 +663,10 @@ class Document (CanvasController):  # TODO: rename to "DocumentController"
         if handler_type == 'mode_class':
             # Transfer control to another mode temporarily.
             assert issubclass(handler, gui.mode.DragMode)
-            mode = handler()
+            if issubclass(handler, gui.mode.OneshotDragMode):
+                mode = handler(temporary_activation=True)
+            else:
+                mode = handler()
             self.modes.push(mode)
             if win is not None:
                 return mode.key_press_cb(win, tdw, event)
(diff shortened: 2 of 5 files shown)


0ion9:
How does this effect the button+modifier way of activating modes? I really don't want to be stuck in a mode after I release the combo that activates it.

jy2wong:
It's supposed to release if you activated it via mouse+modifiers (this is why I kept the movement tool classes as subclasses of `OneshotDragMode`; making them subclasses of `DragMode` made for some odd inconsistent behaviour). I've tested it as thoroughly as I can manage (made sure it behaves sensibly when you activate it with mouse+modifiers, whether you release the modifiers or the mouse button first), though I'm not 100% sure I've thought of everything.

iirelu:
Middle-click dragging leaves you left in pan mode. Aside from that, thanks for trying to implement a few of my suggestions!

achadwick:
Probably want to distinguish temporary (button press) pan/zoom/rotate from permanent (launched from menus/toolbar/keys). Not forgetting that the spacebar thing is pretending to be button 2, sigh.

Perhaps we need to tell newly constructed modes what they were launched from and how? Off the top of my head that's: [pointer buttons configured to switch mode](https://github.com/mypaint/mypaint/blob/496e016846c5019810395aac315cfd72858010e7/gui/document.py#L650) (surprise! [space is a button too!](https://github.com/mypaint/mypaint/blob/496e016846c5019810395aac315cfd72858010e7/gui/document.py#L611)), the regular [activation of a flip action](https://github.com/mypaint/mypaint/blob/496e016846c5019810395aac315cfd72858010e7/gui/document.py#L1742) (surprise! these can be held down to make a mode sticky or not - but this is really just a legacy of the ancient "states" system don't do this), or the regular [activation of a mode radio action](https://github.com/mypaint/mypaint/blob/496e016846c5019810395aac315cfd72858010e7/gui/document.py#L1821) IIRC. Any more?

jy2wong:
Above commit feels pretty hacky, but it seems to be working over here, including the space "button". Test away?

achadwick:
Seems to work OK. I've tested space-is-really-button-2, buttons with and without modifiers, and toolbar/menu activation. The mode stack looks sensible when debugging prints it out.

Now that they're "sticky tools" on the toolbar, I kinda want them to be scrollable too. That should just be a matter of throwing `gui.mode.ScrollableModeMixin` into the recipe though.

jy2wong:
Should be good to go. (Also needed to set `scroll_behavior` to `gui.mode.Behavior.CHANGE_VIEW` from `gui.mode.Behavior.NONE`).

Issue #229, opened 2015-02-28
Right-click menu for brushes


iirelu:
This is flat-out *needed*, and it's kind of silly how we don't have one yet. Obvious settings include "Delete", "Copy", "Edit", "Favourite", but more things will likely be added later.

jy2wong:
So I'm about halfway done implementing this, but figure I can do some code cleanup here too. Before I start, though, is there a good reason for `QuickBrushChooser`s to avoid using the `BrushList` class, @achadwick ?

achadwick:
@jy2wong I can't see any obvious reason, although anything popped up should be kept as a pure single-click chooser with no menus or fancy features IMO. Fancy features like menus can go on `gui.brushselectionwindow.BrushList`.

PR #239, opened 2015-03-05
Add right click menu to brushes

jy2wong:
Starting with add to favourites, remove from favourites, copy, delete
and a placeholder for edit. I also noticed that the recently used list in the toolbar window doesn't respond to right clicks.

menu

This is a WIP that needs more testing. **use a separate throwaway config area for this. It was eating brush groups every launch until I fixed something, and I'm not totally convinced the code is bulletproof yet.**

Partial list of things that I've tried:
* copy brush, restart mypaint
* copy brush, copy its copy, restart mypaint
* copy brush, switch group, switch back, delete copy, restart mypaint
* delete brush, restart mypaint
* favourite brush, restart mypaint
* copy brush, favourite copy, delete copy, restart mypaint

Also would appreciate feedback on the naming of the menu options.

Addresses issue #229 .

### DIFF
--- a/gui/brushmanager.py
+++ b/gui/brushmanager.py
@@ -838,7 +838,7 @@ class BrushManager (object):
         :param group: Name of the group to delete
         :type group: str
 
-        Oprhaned brushes will be placed into `DELETED_BRUSH_GROUP`, which
+        Orphaned brushes will be placed into `DELETED_BRUSH_GROUP`, which
         will be created if necessary.
 
         """
--- a/gui/brushmanager.py
+++ b/gui/brushmanager.py
@@ -38,7 +38,7 @@ PREVIEW_H = 128   #: Height of brush preview images
 
 FOUND_BRUSHES_GROUP = 'lost&found'   #: Orphaned brushes found at startup
 DELETED_BRUSH_GROUP = 'deleted'   #: Orphaned brushes go here after group del
-FAVORITES_BRUSH_GROUP = 'favorites'  #: User's favourites
+FAVORITES_BRUSH_GROUP = u'favorites'  #: User's favourites
 NEW_BRUSH_GROUP = 'new'  #: Home for newly created brushes
 
 ## Internal module constants
--- a/gui/brushselectionwindow.py
+++ b/gui/brushselectionwindow.py
@@ -173,17 +173,19 @@ class BrushPopupMenu(gtk.Menu):
             item.connect("activate", BrushPopupMenu.unfavorite_cb, bl, brush)
             self.append(item)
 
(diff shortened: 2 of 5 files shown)


iirelu:
Thanks for working on this! This is *exactly* the kind of thing that was needed.

I noticed the "Edit brush" option is a placeholder, but already I can tell that it should be renamed to "Edit Brush Settings" for consistency with the Brush > Edit Brush Settings menu option.

"Add brush to favorites" is pretty broken, this is after adding one brush and then another three times:
b

Having the "brush" in all the descriptions is *super* redundant, "Add to favorites", "Remove from favorites", "Copy", "Edit", "Delete" should be used instead. Also, "Copy" usually indicates that something is being copied to the clipboard, which isn't what's happening. It should be renamed to "Clone". 

Again, thanks for starting to work on this!

jy2wong:
Oops. I made sure the favourite callback functions caused a `brushes_changed` event on the right set of brushes, made only one of add/remove show up in the right click menu and renamed everything (in the edit brush dialog, "clone" is called "Copy to New". Should that be changed over to Clone as well?

iirelu:
Thanks for the changes!

Hmm, didn't we used to have a "Trash" folder? I was thinking that instead of an outright delete, there could be a "Move to trash" option, followed by "Delete" and "Restore" when in the trash. Alternatively, we could add a confirmation dialog to the "Delete" button, as I can see that becoming a big issue for some.

Apart from that, this looks good from a functionality perspective! @achadwick, what do you think of the code?

jy2wong:
Just as a sanity check, it's sensible to delete a brush from both favourites and the brush group it was in, yes?

iirelu:
@jy2wong Hm, not really. "Clone" should probably go too, for brushes in the favourites. If a user wants to clone or delete something, they can do it from the original.

To the user, the favourites is a special folder that doesn't actually contain brushes, just links to brushes. Of course, it doesn't actually work that way, but that's the most comfortable way to picture it.

jy2wong:
Yeah okay, sounds reasonable.

Also, thanks for putting up with all my silly UI questions. I've been using linux for years and have a...very high tolerance for bad UI decisions. :sweat:

iirelu:
It's okay! Asking a silly question is better than making a silly mistake.

jy2wong:
Left a86ebfe as a separate commit just to make it clear what's been changed; I'll smush it together with the main commit once everyone's happy.

iirelu:
Frankly I much prefer a detailed commit history to a "clean" one, so you don't have to rebase.

Tested some more and this works great! Thanks for putting up with all my nagging, will merge this now!

PR #248, opened 2015-03-09
removed an optimization causing a loss of data when exporting as png

aigleghis:
If a layer contains really few strokes (like a single short line), when exporting as multiple png, this layer's png will be empty.

Repro steps:
1: create a new document.
2: activate the Frame to have a finite canvas size.
3: draw a short line in the middle of the canvas.
4: export as multiple png.
5: notice the png corresponding to your layer is empty.

### DIFF
--- a/lib/tiledsurface.py
+++ b/lib/tiledsurface.py
@@ -506,8 +506,6 @@ class MyPaintSurface (object):
         if 'alpha' not in kwargs:
             kwargs['alpha'] = True
 
-        if len(self.tiledict) == 1:
-            kwargs['single_tile_pattern'] = True
         pixbufsurface.save_as_png(self, filename, *args, **kwargs)
 
     def get_tiles(self):


achadwick:
Ummm - this might flag something more worrying. Shouldn't the pending changes be flushed out to the doc in this case, meaning you'll have an up-to-date backing surface?

https://github.com/mypaint/mypaint/blob/db06f71a145ca0e4f6c2568275547e677262e8b3/lib/document.py#L471

We're supposed to call that before doing things like saving so that the ink and freehand tools have a chance to push their pending stuff to the doc.

aigleghis:
actually,  it doesn't seem to be caused by unseen pending changes.   The issue occurs even when I load a document and try to export layers from it.

achadwick:
Oof. Presumably one with a single-tile tilemap?

achadwick:
Er, can't reproduce with the following, and I've verified that it's applying the `single_tile_pattern` optimization. The .XXX.png corresponding the the icon layer is not empty after saving: it contains a 48x48px transparent-bg copy of the original layer data.

ORA attachment: mypaint_icon.ora

(It's not really a .png, forgive the broken image. just tricking the Github CMS into making it an attachment. Rename to .ora and open!)

Does it only kick in with "export" and not "save" on a doc like this?

achadwick:
Nope, export works fine too.

It's also not sensitive to "Show background" being turned off, or big backgrounds...

aigleghis:
Actually,  the bug occurs only when exporting as "multiple transparent png".  The produced image.000.png is empty.  I've put some log in the code to see what was happening;  it is skipping rendering because the flag single_tile_pattern is true.  I'm sorry I can't send my .ora file right now but I'll do as soon as I get home.

aigleghis:
I cannot reproduce the issue with the ora file you provided me, the export works well.  However, when I activate and resize the Frame to a bigger size (in a way that the icon is centered), the next "export as Multiple PNG transparent" has the issue.  The issue seems to get fixed by itself once you save the .ora though.

I'll send you a ora file which fails 100% of the time (only when exporting as multiple PNG transparent): 
test ora

achadwick:
@aigleghis: sorry for the delay with this. I was able to reproduce it with your test image for the case you mentioned, and I've pushed a variation on your patch that restricts it to just the case we're trying to optimize. It seems to fix the data loss in your test,ora above: are you able to re-test?

Thanks for the contribution!

Issue #232, opened 2015-03-03
Exception dialog enhancements (grab-bag): triple backticks, autoselect, github link?


achadwick:
Fallout from #230 here. I've been thinking about what could be done to make the "crash dialog" more useful. What do people think about:

1. Adding triple backtick lines before and after the formatted backtrace to make formatting in github nicer
2. Adding a GtkLinkButton pointing at https://github.com/mypaint/mypaint/issues to let them report it?
  * Or better still a [Search for this error](https://github.com/mypaint/mypaint/search?utf8=%E2%9C%93&q=NameError%3A+global+name+%27PaintingLayer%27+is+not+defined&type=Issues) button, FIRST, which only unlocks a SECOND [Report this Error](https://github.com/mypaint/mypaint/issues/new?title=foo&body=bar) button when pressed...
  - It looks as if the title and body can be pre-filled... :fearful:
3. Automatically selecting the Details... text when GtkExpander is opened (or perhaps via a button in the details section)

Testers quite frequently just copy and paste what they see in the "crash" dialog into a new issue. Perhaps we could smooth that process?

Some downsides to smoothing things like this: people aren't going to tell us what they were trying to do when they encountered the exception, they're just going to click on Submit. Is there any way of forcing/interlocking that behaviour too?

achadwick:
Re search buttons: we'd use the raw exception class and error text. That seems to have good results in github and flags up what you'd expect (see links above).

Prior art: Debian's ancient "reportbug" tool makes searching the database a requirement before submitting. It seems to do a good job there, from experience.

odysseywestra:
I think the first step would be to first add the three backtips to the backtrace window. At least that will help with the correct formatting. Plus instead of directly linking to the issue page, maybe link to `contribute.md` for instructions on how to submit reports. That will give new users an idea how to report bugs and keeping it simple.

jy2wong:
Maybe prefill the body with a template, e.g.
```
<REPLACE ME WITH A DESCRIPTION OF THE BUG>

Steps to reproduce:
<GIVE STEP BY STEP INSTRUCTIONS ON HOW TO REPRODUCE THE BUG>

<automatically put crash log here>
```
as a compromise between making it easy to report issues and getting meaningful bug reports?

achadwick:
Already closed via the referenced merge.

PR #257, opened 2015-03-12
Better crash dialog

jy2wong:
I guess the new buttons could also go right above details (and would probably help with the peculiar linewrapping), but I feel like that's maybe kind of weird for a dialog window.

It looks like someone intended the trace to be printed out in monospace, but it wasn't doing that for me. Fixed for me now, though I can't tell if I've now broken it for other versions of...pygtk? pango?

A first stab at #232.

### DIFF
--- a/gui/gtkexcepthook.py
+++ b/gui/gtkexcepthook.py
@@ -24,6 +24,7 @@ import sys
 import traceback
 from cStringIO import StringIO
 from gettext import gettext as _
+from urllib import quote_plus
 
 import gtk2compat
 import gtk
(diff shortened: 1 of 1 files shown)


iirelu:
Before:
a
After:
b

Looks interesting! How does the "Report issue" button fare with long exceptions?

There's also some weirdness with newlines for the default text:
b

achadwick:
This is great! I'm wondering whether we should have a general `report_bug()` function (and "Report a Bug" menu item) as well?

Maybe use something like

```python
body = """\
    Long...

    ... multi-paragraph, and
    line wrapped
    body text here.

    It can be
        indented
    too!
"""
body = "\n\n".join([
    "".join(textwrap.wrap(p, sys.maxint))
    for p in textwrap.dedent(body).split("\n\n")
])
```

for the newline thing?

PR #280, opened 2015-03-30
Updated slovak translation

prescott66:


achadwick:
Looking good. Thanks for testing your the new C_() stuff, and I'm glad it's working for disambiguating uses. I'll be sure to use it for other possible clashes.

Issue #181, opened 2015-01-24
Clipping group


loentar:
Please implement clipping group feature similar to SAI. This feature is extremely useful when shading.

```
  [ [v] some other shadow ] - clipped to "some other layer"
  [ [ ] some other base ]
  [ [v] highlight ]  - clipped to "base"
  [ [v] shadow-dark ] - clipped to "base"
  [ [v] shadow ] - clipped to "base"
  [ [ ] base ]
```

All immediate layers which is above of base with "clipping group" check enabled will be clipped to "base" layer.

Currently this feature can be done using masks, but you must have two base layers (one for mask) which is inconvenient:

```
[ Group ]
    + copy of base - mask. Mode: Destination In
    + shadow. Mode: Normal/Darken/etc...
    + base. Mode: Normal
```

And to reproduce the first scheme you must have a mask for each layer:

```
[ Group ]
    + copy of some other base - mask. Mode: Destination In
    + some other shadow. Mode: Normal/Darken/etc...
    + some other base. Mode: Normal
    + copy of base - mask. Mode: Destination In
    + highlight. Mode: Normal/Darken/etc...
    + copy of base - mask. Mode: Destination In
    + shadow-dark. Mode: Normal/Darken/etc...
    + copy of base - mask. Mode: Destination In
    + shadow. Mode: Normal/Darken/etc...
    + base. Mode: Normal
```

It's too complicated..

achadwick:
This sounds way too complicated to do for 1.2. Postponing.

0ion9:
@loentar I think you make this sound more complicated than it is, because in line one of your first list you refer to the unlisted layer "some other layer". I think you may have meant "some other base".

Overall I would say your proposal needs more specifics. For example, is the clipping mask used supposed to clip to the next layer below it in the stack that has a 'destination in' mode, counting only the current group, not parent groups etc?  Then it ought to be relatively simple to introduce a layer of code that filters such a layerstack into the final layerstack that needs to be rendered.
OTOH, if the behaviour can transcend the current group level, then you definitely need to demonstrate how it should behave in a greater variety of cases.

loentar:
Oh, yes, sorry for typo. I meant "some other base". Of course all the layers in the layer group is clipped ONLY to the one base layer which is below of it. Only the layers in the group is clipped, I don't expect any other interaction between layers. "base" and "some other base" is a just two independent layers like without shading, they don't interact unless you change the Normal mode.

[...]

PR #284, opened 2015-04-06
Implemented new blending modes "Destination Atop" and "Source Atop"

loentar:
Please review my changes for https://github.com/mypaint/mypaint/issues/181.

### DIFF
--- a/lib/blending.hpp
+++ b/lib/blending.hpp
@@ -107,6 +107,32 @@ class BufferCombineFunc <DSTALPHA, BUFSIZE, BlendNormal, CompositeDestinationOut
 // Da'  = Da × (1 - Sa)
 
 
+template <bool DSTALPHA, unsigned int BUFSIZE>
+class BufferCombineFunc <DSTALPHA, BUFSIZE, BlendNormal, CompositeDestinationAtop>
+{
+    // Partial specialization for normal lainting layers,
+    // working in premultiplied alpha for speed.
+  public:
+    inline void operator() (const fix15_short_t * const src,
+                            fix15_short_t * const dst,
+                            const fix15_short_t opac) const
+    {
+        for (unsigned int i=0; i<BUFSIZE; i+=4) {
+            const fix15_t as = fix15_mul(src[i+3], opac);
+            const fix15_t one_minus_ab = fix15_one - dst[i+3];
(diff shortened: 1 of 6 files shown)


achadwick:
This needs to be discussed on the CREATE list first, and the [OpenRaster specification](https://wiki.freedesktop.org/www/Specifications/OpenRaster/Draft/LayersStack/) doesn't discuss it.

I think we should probably add src-atop too, for symmetry :D

I don't think this should be merged until we have buy-in from the Krita people at least.

achadwick:
I'll check over the maths when localtime >> 03:00. Looks basically sound, and `tests/test_compositeops.py` likes it, but let's get some conversation happening elsewhere too.

achadwick:
I've finally had a chance to look over this, and work through the maths. It's a bit too fringe-y right now:

atop_bad

what I expected is:

atop_good

which looks loads cleaner.

I think I've tracked it down to the fact that the code in this branch assumes that the buffers it's operating on contain unmodified RGBA - they don't, each pixel's RGB values are premultiplied by the pixel's A, so when you see 

> co = as * Cs * ab + ab * Cb * (1 - as)

in the [w3c spec](http://www.w3.org/TR/compositing-1/#porterduffcompositingoperators_srcatop), treat it as (substituting wildly):

> co = src[n] * ab + dst[n] * (1 - as)

I've fixed up the template specialization implementation, and I'll fix up the unspecialized implementations too in case we need them sometime (perhaps in test code for use as a reference spec?)

achadwick:
Here's the test image I used:

atop1.ora

It's not a .png, so download it and give it a .ora to experiment with it!

achadwick:
Incidentally, the alpha expressions simplify in a way that I guess you'd expect:

```
6-> ao = as*ab + ab*(1-as)    #src-atop
#7: ao = (as*ab) + (ab*(1 - as))
7-> simplify
#7: ao = ab
```

```
5-> ao = as*(1-ab) + ab*as    #dst-atop
#6: ao = (as*(1 - ab)) + (ab*as)
6-> simplify
#6: ao = as
```

That's using mathomatic because lazy, but pencil+paper agree :D

achadwick:
Merged with 48f483dadeb1ede0cf7fa44b14d180f055f7a27e and 0ac6fd933358feb3e2bbb89de3c65d76ddb4da2c.

PR #302, opened 2015-04-26
Brushes moved to a separate mypaint-data module.

Jehan:
For GIMP, where we now have a usage of libmypaint in the dev branch, we can just guess the installation prefix of mypaint. But except for common prefixes (/usr/ or /usr/local) or else try the same prefix as GIMP, we may not be able to guess the right prefix where mypaint got installed.

So we propose a separate module for data (which would at least contain the brushes, and any other relevant data) into their own repository, and add a pkg-config to this data repo with a proper pc file allowing third party applications to discover mypaint data location (using custom pkg-config variable)?

I have initiated a repository containing current brushes/ (as is, just moved from the current mypaint repo), a scons build environment, and a pkg-config generated with proper datadir information.
The repository can be cloned from git@libreart.info:/home/git/mypaint-data.git
If you are happy with it, I would invite the mypaint project to move it as an official repository.
The current merge request is for having this repository as a submodule of the main mypaint repo (of course the submodule url has to be changed before actual merge happen with the finale official url).

I can make any change necessary, for instance if you want me to move other data into this data module (palettes/ or whatever you want). I moved only the brushes/ in this first version since that's the only data our third party program seem to really be able to access currently.
Thanks.


achadwick:
Got HTTPS access or anon access credential for that repo? I can't see it - I'm not a LILA member.

Name is great, slots in nicely with the debian package's mypaint-data-extras :D

Two organizational concerns prevent me from doing anything on this right now:

1. It is too close to the 1.2 release for MyPaint to make any changes in this direction immediately.
2. The primary repository must be hosted on Github, alongside mypaint. I don't think this is even negotiable for submodules; our users need to be able to access the repository. I'm really cool with giving teams read/write access to individual repos though!

Jehan:
Hi,

I'm so sorry about the non public repo url! I did this yesterday quite late
and tired (in Toronto for LGM) and obviously didn't think it through. I
made you a public github repo instead: https://github.com/Jehan/mypaint-data

About the 2 organizational concerns:
1. No problem, it can be after 1.2. We are not in extra hurry since it only
affects the GIMP master branch (not our stable release).
2. Solved above. Also even though I made the original repo, I was obviously
intending it to be taken as a "mypaint team" primary repo, and not under my
name! I just boostrap the first version, that's it. :-)

Also a note of my own: I was unsure of the licensing for current mypaint
brushes. So I just copied over from mypaint repo COPYING and parts of the
LICENSE file which look like they pertains to brushes. Yet the only clear
licensing statement I can read there is about "The brushes and their icons
from Set #2". For the other brushes, I am completely unsure if they are not
under the GPL v2 license like the code, even though you also say "We
currently require Public Domain licensing, or CC0, for brushes […]", in
which case I imagine all brushes would actually be CC0/Public Domain?

Thanks!

Jehan

achadwick:
Also relevant, and may end up being done at the same time: mypaint/mypaint#535

Jehan:
Are you planning to do this soon? If so, I can update my repos for this to happen easily.
Some things have happen in GIMP master branch recently and the mypaint brush feature have been moved out from playground to actual feature just today. So if we could also have proper data package to link to, it would be perfect. :-)

achadwick:
Not until the mypaint 1.2.0 release - but that's just around the corner. Hopefully by the end of the year, then I'll take another look at this and decide how/what/when to merge. Sorry for the delay!

Also relevant to splits and what GIMP wants out of libmypaint: mypaint/libmypaint#28, mypaint/libmypaint#29.

Jehan:
I updated my repo, which closed the pull request and I can't reopen it apparently. I reopened a new pull request #538 instead.

PR #327, opened 2015-05-26
Found the tooltips applied to the wrong places in HSV Square

ShadowKyogre:
This one's just a minor fix for swapping the tooltip labels around in HSV Square. ^^;


iirelu:
Haha, not the worst bug in the world to deal with! Thanks.

Issue #344, opened 2015-06-14
Moving the mouse and simultaneously moving the tablet pen causes an OverflowError


aaronjwoodard:
Here's the error:
```python
Exception while analyzing the exception.
Traceback (most recent call last):
  File "/usr/local/share/mypaint/gui/freehand.py", line 555, in _motion_queue_idle_cb
    self._process_queued_event(tdw, event)
  File "/usr/local/share/mypaint/gui/freehand.py", line 597, in _process_queued_event
    self.stroke_to(model, dtime, x, y, pressure, xtilt, ytilt)
  File "/usr/local/share/mypaint/gui/mode.py", line 760, in stroke_to
    cmd.stroke_to(dtime, x, y, pressure, xtilt, ytilt)
  File "/usr/local/share/mypaint/lib/command.py", line 389, in stroke_to
    xtilt, ytilt, dtime,
  File "/usr/local/share/mypaint/lib/layer/data.py", line 1319, in stroke_to
    pressure, xtilt, ytilt, dtime
  File "/usr/local/share/mypaint/lib/mypaintlib.py", line 285, in stroke_to
    def stroke_to(self, *args): return _mypaintlib.PythonBrush_stroke_to(self, *args)
OverflowError: in method 'PythonBrush_stroke_to', argument 7 of type 'float'

```
It happened to me when I was using my tablet and bumped into my mouse. It seems that MyPaint doesn't like it when you move both at the same time. The result is MyPaint will not let you continue to draw.

OS: Debian (Linux 3.16) amd64
Tablet: Wacom Inutos

achadwick:
Interesting. When you dismiss the exception dialog, will MyPaint let you continue to draw?

Can you provide more info on your tablet model, and whether you were pressing a mouse button at the time? Basically let me know how to replicate this exception here.

martinxyz:
Uh, "Exception while analyzing the exception", which is why we can't see the values. But it smells like infinity or NaN... maybe some mouse event delivered with an earlier/same timestamp compared to a tablet event? I can't reproduce (Debian testing, Linux 3.16, Intuos3, xfce). Try to enable the "input test" window in Help->Debug, maybe you'll get more info.

achadwick:
Oh yes. In which case the dialog will be a bit useless probably.

I can't reproduce this here with my Intuos5 + laptop touchpad setup.

[...]

PR #345, opened 2015-06-17
Properly emulate old scrolling functionality

iirelu:
The logic for the old step-wise scrolling functionality was improporly recreated in the new smooth-scrolling-based logic as of 189e195. This makes it *exactly* the same as it was before.

I'm not familiar with this code so there might be some subtleties I'm missing that have to be accounted for, in which case just tell me and I'll fix it.

Right now it seems that this is only used if you untick the "Handle smooth scrolling events when devices send them" in your preferences, so I think there's still problems with the code not properly figuring out what's actually a smooth scrolling event.

### DIFF
--- a/gui/mode.py
+++ b/gui/mode.py
@@ -473,12 +473,10 @@ class ScrollableModeMixin (InteractionMode):
                 if constrain_smooth:
                     # Needs to work in an identical fashion to old-style
                     # zooming.
-                    while self.__total_dy > 1:
-                        self.__total_dy -= 1.0
-                        doc.zoom(doc.ZOOM_OUTWARDS)
-                    while self.__total_dy < -1:
-                        self.__total_dy += 1.0
+                    if direction == gdk.SCROLL_UP:
                         doc.zoom(doc.ZOOM_INWARDS)
+                    elif direction == gdk.SCROLL_DOWN:
+                        doc.zoom(doc.ZOOM_OUTWARDS)
                 else:
                     # Smooth scroll zooming is intended to resemble what
                     # gui.viewmanip.ZoomViewMode does, minus the


achadwick:
Bear in mind that the "Handle smooth scrolling events when devices send them" setting is one that requires MyPaint to be restarted.

achadwick:
@iirelu You're seeing a lot of weirdness that I'm not. I'm wondering if your device is sending a mixed stream of smooth and non-smooth events for the same wheel rotation. Maybe that's just lingering worry from dealing with #344 though.

Could you stick a few debug statements in there (on master) to rule that out please?

```patch
diff --git a/gui/mode.py b/gui/mode.py
index 3f4149f..a26bc3e 100644
--- a/gui/mode.py
+++ b/gui/mode.py
@@ -470,6 +470,10 @@ class ScrollableModeMixin (InteractionMode):
         # We don't rotate any more though. Srsly, that was awful.
         if scroll_action == gui.device.ScrollAction.ZOOM:
             if direction == gdk.SCROLL_SMOOTH:
+                logger.debug(
+                    "PR#345: zoom-SMOOTH dx=%r,dy=%r",
+                    event.delta_x, event.delta_y,
+                )
                 if constrain_smooth:
                     # Needs to work in an identical fashion to old-style
                     # zooming.
@@ -501,9 +505,11 @@ class ScrollableModeMixin (InteractionMode):
                     self.__reset_delta_totals()
             # Old-style zooming
             elif direction == gdk.SCROLL_UP:
+                logger.debug("PR#345: zoom-nonsmooth-INWARDS")
                 doc.zoom(doc.ZOOM_INWARDS)
                 self.__reset_delta_totals()
             elif direction == gdk.SCROLL_DOWN:
+                logger.debug("PR#345: zoom-nonsmooth-OUTWARDS")
                 doc.zoom(doc.ZOOM_OUTWARDS)
                 self.__reset_delta_totals()
 ```

And run with

```$ MYPAINT_DEBUG=1 ./mypaint -c /tmp/cfgtmp1 2>&1 | grep 'PR#345'```

There should be no interleaving of the "zoom-nonsmooth" and "zoom-SMOOTH" messages at any time, even when "Handle smooth scrolling events when devices send them" is set to "When devices send them" (and after MyPaint is restarted if you change it).  If they interleave, we need to redo this bit of code pronto.

I'd also like to see what the dx and dy values your device reports for the smooth flavour of events are. For this scrollwheel mouse here, it's ±1.0 exactly for each wheel notch.

iirelu:
> Bear in mind that the "Handle smooth scrolling events when devices send them" setting is one that requires MyPaint to be restarted.

...Ah. Time to close this PR and open a... Different one.

PR #367, opened 2015-07-14
Use Travis' container-based infrastructure.

QuLogic:
Now that Travis has accepted all my requests to whitelist mypaint's dependencies, I think this should work. Container builds start and run a little bit faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,27 @@
 language: cpp
 compiler:
     - gcc
-install:
-    - sudo apt-get update -qq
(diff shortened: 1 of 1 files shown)


achadwick:
It's missing the `git submodule update --init` line, which will cause the build to fail.

QuLogic:
Nope, the build [passed](https://travis-ci.org/mypaint/mypaint/builds/70832908). (Why is Travis not posting status here?)

Travis already does a [`git submodule init`](https://travis-ci.org/mypaint/mypaint/builds/70832908#L83) and [`git submodule update`](https://travis-ci.org/mypaint/mypaint/builds/70832908#L85). In the original non-container builds, that line [did nothing](https://travis-ci.org/mypaint/mypaint/builds/70749618#L139).

achadwick:
Yep, that works fine. My apologies, and thanks for the update!

Issue #412, opened 2015-08-20
Active tab not saved in sidebars and floating dockpanel stacks


zb13y:
MyPaint opens with last tab active. I think that first tab must be active and mypaint doesn't save which tab is active. Krita and Gimp saves active tab. Windows 10 x64, MyPaint 1.2.0 beta 1.

![213](https://cloud.githubusercontent.com/assets/13454470/9381553/e8f336ea-4751-11e5-9f0e-7771d4b2ae91.jpg)

achadwick:
I think the last tab should be active, in the absence of better information :grinning: but that's only a difference of opinion.

It would be a good improvement to save which tab was last active in each notebook. Let's do that.

achadwick:
I've changed the title because this affects all builds of MyPaint 1.2.0-beta. Bitesize because it's a small task: the code can be found in `gui/workspace.py`, and I'd *love* patches or pull requests for this sort of thing.

PR #413, opened 2015-08-22
Remember last active tab in each notebook

iSoron:
This closes mypaint/mypaint#412


achadwick:
Looks good! Thanks for the fix ☺

Issue #414, opened 2015-08-23
Where did 'Hide Layer' and 'Lock Layer' go?


ydahhrk:
Is this normal?

![Iconless Layers palette](https://dl.dropboxusercontent.com/u/95836775/MyPaint/LayersNoIcons.png)

I recall when I upgraded to <del>1.2</del> beta 1.2 the eye (Hide/show layer) and lock (prevent edits) icons were moved to the <del>right</del> left. At some point then they randomly disappeared. I don't know how to retrieve them :(
(Renaming the ~/.mypaint folder doesn't do anything.)

I was using the 1.2.0-beta.1+gitexport.a7012b4 version. I upgraded to the current master (dc79f2564e) and the problem persists.

Ubuntu 14.04.

achadwick:
(Version 1.2 has not been released yet - we're still in beta)

No, that's not normal. There has been one recent change: eye and lock should be on the left now, and type icons on the right (after #144 was closed):

![flippedlayout](https://cloud.githubusercontent.com/assets/61299/9093288/20311146-3ba3-11e5-94ec-bf64b4458eb1.png)

I did some minor work on the column widths as part of it. Perhaps it's related. Can you investigate whether reinstating any of the lines deleted by 6048998bf56d6260af95c24b8fcdc26a5e5f5df5 fixes the problem?

Quick question: do you get the folder icon appearing on the right if you make a group?

ydahhrk:
> (Version 1.2 has not been released yet - we're still in beta)

I know :)

> eye and lock should be on the left now

Oops, yes, that's what I meant :)
(see the scratched text in my original post.)

> Can you investigate whether reinstating any of the lines deleted by 6048998 fixes the problem?

Ok, working.

> Quick question: do you get the folder icon appearing on the right if you make a group?

No.

achadwick:
(bit in brackets was mostly for future searchers - sorry ☺)

I can't reproduce this in a VM with MyPaint master (running in-tree, with a fresh config) and Ubuntu 14.04 LTS. As part of my testing, I upgraded from GTK 3.10.8-0ubuntu1.5 to GTK 3.10.8-0ubuntu1.6.

It should look like

![image](https://cloud.githubusercontent.com/assets/61299/9428722/fc073f1e-49af-11e5-956b-16596238c0b7.png)

Do you get any difference when running with 

    ./mypaint -c /tmp/mypaint-confarea-throwaway-1

from within the source tree?

[...]

PR #415, opened 2015-08-23
layers panel: add explicit width constraint to columns

ydahhrk:
Otherwise unknown causes can make them disappear.
The columns are "Visibility", "Lock" and "Type".

Addresses mypaint/mypaint#414.


achadwick:
Thanks for investigating and fixing!

achadwick:
(Er, ignore that reference - the commit is irrelevant. Sorry.)

Issue #253, opened 2015-03-12
Color picker awkwardness when invoked via keyboard


0ion9:
I have the right arrow on the numpad bound to Color Picker.
Its behaviour has changed undesirably, recently:

* Previously, I could pick up a color just by hovering over what I wanted and hitting KP_6 - the color is instantly picked up.

* Now, I have to first hold (not press) KP_6, then click on a color to pick it up. Further, as soon as I take the pen away, I must then release KP_6 and re-press it in order to pick a color again.

iirelu:
By previously do you mean 1.1.0 or an earlier version? If so, that likely changed a long time ago. I can't seem to produce your second bullet point, though. With ctrl (the default eyedropper button), I can select multiple colours while its still held.

0ion9:
@iirelu : CTRL+clicking to pick works for me (that is, I can do it. I don't think it's as good as hover picking.).

However, that is a **button** assignment, not a key assignment.
The type of thing I am talking about is a key assignment:

![screenshot_480](https://cloud.githubusercontent.com/assets/115782/6613003/3c21eb40-c8d6-11e4-9db4-676b3574973e.png)

achadwick:
@0ion9 Does it change either side of 8f438497c0767b913649fb5cfdb97545ae769d2a (/guess)? If it's not that, would you mind triaging for a bit to track down where this started - or give us some versions to work with so we can do it ourselves?

That's normally bound to the "R" key. Is the behaviour different with the defaults?

0ion9:
@achadwick 
* Behaviour is the same with the defaults.
* The parent of 8f43849 does not exhibit the misbehaviour.
* 8f43849 exhibits the misbehaviour

So it's 8f43849, all right.

achadwick:
Reproduced with 08f3cc53133f50eca8d9211d6b267a8d971081c5. Thanks for the report.

achadwick:
Another hopefully quite small task here.

achadwick:
**Triage note:** reproducible, no string changes needed, not super-urgent; bump to the beta phase for fixing.

borda-flash:
I will try on this

achadwick:
@borda-flash Any progress on this?

iisupreme:
Hi @achadwick 

Is the update required to reverse the changes made in 8f43849...

Regards
Val

PR #416, opened 2015-08-23
Restore old color picker behaviour (hover to pick)

iSoron:
In older versions of mypaint, it was possible to pick a color by simply
holding down the 'pick color' hotkey and hovering over a color. At some
point, this changed. Users were now required to hold down the hotkey and
then drag over the color. This patch restores the old behaviour.

Resolves mypaint/mypaint#253.


achadwick:
Many thanks for the patch :grinning:

iSoron:
Thanks for merging it!

Issue #244, opened 2015-03-08
Scratchpad detached tab


americogobbo:
When you've the tab detached of normal position, for example, you put over canvas or another screen, is not possible to move the scratchpad canvas with 'space+mouse cursor'. This behavior is normal?

achadwick:
What OS, desktop environment, version of MyPaint?

americogobbo:
Ubuntu Gnome 14.04 and the last git master release: 1.1.1-alpha.20150306+git.6b3a1e7

achadwick:
Thanks!

Reproduced with {Debian testing+unstable (2015-03-08), GNOME 3.14, git master at 6b3a1e7f21e6cc597baa31a48f6947ced9f760e3 as above}.

If I detach the scratchpad, holding Space and *just* moving the mouse (no click) sometimes works, and sometimes doesn't. When it works in this config, it doesn't cancel properly when the space bar is released: you need to click to end the drag.

Docking the scratchpad restores the expected behaviour.

americogobbo:
OK, works.
The features on bottom toolbar are working too, but they appear as disable (tab not focused) ... but this is a behavior of all detached tabs... I've noted now... is normal? or I need open another issue for it?

achadwick:
That's a consequence of the window not accepting keyboard focus (with some gtk3 themes?). Raise it as a separate cosmetic issue please: annotated screenshots of the windows will help clarify the problem to other readers!

americogobbo:
I've made a small video to show this behavior:
https://www.youtube.com/watch?v=L8rthkYOnjs
When you use space key to drag the scratchpad canvas focused it don't works... but if you click on mypaint canvas is possible to drag the scratchpad canvas.

PR #417, opened 2015-08-24
Make floating windows process keyboard events

iSoron:
Before this patch, some events were only processed if sent to the main
drawing window. Keyboard events sent to floating windows were silently
ignored. This caused strange behaviour when trying to pan a detached
scratchpad using spacebar + mouse drag.

Resolves mypaint/mypaint#244.

### DIFF
--- a/gui/workspace.py
+++ b/gui/workspace.py
@@ -1848,6 +1848,8 @@ class ToolStackWindow (Gtk.Window):
             win.set_decorations(decor)
             wmfuncs = Gdk.WMFunction.RESIZE | Gdk.WMFunction.MOVE
             win.set_functions(wmfuncs)
+            self.connect('key-press-event', toplevel.key_press_event_cb)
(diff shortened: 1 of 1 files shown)


achadwick:
Oh goodness, our keyboard dispatch code is a mess.

I don't think it's the right approach for the workspace manager to know that it's packed inside a window that can do fallback dispatch like this. The Workspace object should be ignorant of that sort of thing.

The extra key handlers can be hooked up from outside the workspace code, in a handler for the `workspace.floating_window_created` `@event`. And there's one such place already, set up to register the new window with `app.kbm` (the `gui.keyboard.KeyboardManager` instance for the app).

I think it's best to redo the way the main window's keyboard handling works: better to use the dedicated kbm and dispatch it explicitly than faff around with events falling through to other events.

I'll rework this (I want to do a little refactor in this area anyway) but thank you for showing me where to start looking on this issue!

Issue #272, opened 2015-03-27
Eyedropper tool doesn't work on the scratchpad


AtsusaKaneytza:
Version: MyPaint 32-bit 1.1.1-alpha.20150319+git.4d5bd9b
Windows 8, 64-bit
It used to, but now it doesn't. Using the keyboard shortcut or clicking the eyedropper tool and trying to select the color doesn't work.

artsnake:
There are two eyedropper shortcuts one using the R key to select what the cursor is hovering over. The other uses the ctrl key plus the touch of the pen on a tablet or mouse click.
On my Ubuntu 14.04 box the ctrl plus click works on the scratchpad. The R key does not. 
I think that works just fine.

iirelu:
Still a bug, though.

jplloyd:
This was fixed by the 1.2 release (at the latest). Tested on Win7, 10 and Ubuntu.

PR #418, opened 2015-08-25
Dispatch mode-flip actions to the active document

iSoron:
All mode-flip actions are currently being sent to the main document. This
causes issues such as mypaint/mypaint#272. It also makes it impossible to
use other tools (such as line-drawing tools) on the scratchpad. With this
patch, mode-flip actions are sent to the active document instead.

Note that this does not completely solve issue mypaint/mypaint#272, since
it is still impossible to use the eyedropper on the scratchpad by clicking on
the toolbar.

### DIFF
--- a/gui/document.py
+++ b/gui/document.py
@@ -1810,6 +1810,11 @@ class Document (CanvasController):  # TODO: rename to "DocumentController"
         when entered.  See also: `gui.mode.DragMode`.
 
         """
+        # If this is not the active document, dispatch the action to it.
+        active_doc = Document.get_active_instance()
+        if active_doc != self:
+            return active_doc.mode_flip_action_activated_cb(flip_action)
+
         flip_action_name = flip_action.get_name()
         assert flip_action_name.startswith("Flip")
         # Find the corresponding gtk.RadioAction


achadwick:
Ah, bother. I think my fix for #244 / refactor inspired by #417 may have broken this: _get_active_doc() is gone (and it was internal anyway). The active doc can be got from `Document.get_active_instance()` though - and that's not going away any time soon:grinning:

Would you be able to patch up this branch and rebase its against master?

iSoron:
Sure. I will have a look.

achadwick:
Alternatively, mode_flip_action_activated_cb() could be moved to either the Application singleton or its DrawWindow. Then it won't look so weird to have "is this the right one? no? okay, that one over there..." code.

(Arguably that's where the modes stack lives too though, so maybe we should keep this patch small and simple for now, since there's a lot of refactoring paralysis this could embark us on...)

iSoron:
Yes, that would make a lot more sense. Feel free to write that patch and close this pull request if you would like.

iSoron:
Updated the patch. It now uses Document.get_active_instance().

achadwick:
Thanks! It's probably best to leave the heavy rearchitecting till after the 1.2.0 release :neutral_face:

Issue #432, opened 2015-10-03
Reverse color axis direction for 2 color pickers


BlenderNPR:
While producing a comprehensive color course, we discovered reverse color axis in MyPaint **1.20**

The problem is the axis of some color pickers are reverse in direction to sliders version.
![screenshot from 2015-10-02 23 46 26](https://cloud.githubusercontent.com/assets/11452632/10263932/275c74f8-6a31-11e5-9c46-b27841e346f6.png)
![screenshot from 2015-10-02 23 46 29](https://cloud.githubusercontent.com/assets/11452632/10263939/3c3842bc-6a31-11e5-8f2e-b578a75e6124.png)
While the sliders do it correctly.
![screenshot from 2015-10-02 23 46 32](https://cloud.githubusercontent.com/assets/11452632/10263941/42f08f4c-6a31-11e5-8185-af912cae8a66.png)

Blender's color pickers follow the axis direction rule strictly.
Increase always toward the top or right direction.
![sv hsquare](https://cloud.githubusercontent.com/assets/11452632/10263956/7ce30d10-6a31-11e5-919f-35c4d76320ce.png)

GIMP's color pickers on the other hand have more serious UX problems, but that's for another bug report else where.

BlenderNPR:
I missed 1 color picker UX inconsistency.

Color wheel, clockwise or CCW?

**Note:** Clockwise is referencing 0 deg toward 360 deg (the longer path)
red to yellow in clockwise direction

On Triangle Color wheel, hue is CCW. (incorrect)
On Square Color wheel, hue is CW. (correct)

For consistency, please use CW direction. Blender also using CW direction with red starting at the bottom (0 deg). The position of red hue isn't a problem (each software has its own convention), but the direction to next hue is important.

odysseywestra:
For semantics sake, could you put down which version of MyPaint you are
using and your operating system. You can easily get that information by
going to Help > Debug > Stimulate a Crash and then click on Details. Thanks.

achadwick:
I can't do anything about the triangle one: it's a GTK builtin: [GtkHSV](https://developer.gnome.org/gtk3/stable/GtkHSV.html), so its direction is fixed. Sorry! Not only is the Hue direction fixed to CCW, it doesn't respond to the colour wheel type setting
GtkHSV is old and deprecated upstream, and Should Be Removed from MyPaint because it has a ton of old bugs relating to it: see #268 about replacing it, and #266, #169 too.
[...]

PR #436, opened 2015-10-03
Do not adjust amt or the parameter f1 by flipping it around. 

ShadowKyogre:
This is inconsistent with the ideal examples provided by @BlenderNPR
on #432. On a side note, there should be comments on what f1 and f2 __are__.


achadwick:
Patch looks fine. Don't suppose you could edit the commit message of the first commit (currently 2819b318e50496b555bd1c59b7459b5a051f8dac) to follow the [usual](http://365git.tumblr.com/post/3308646748/writing-git-commit-messages) [conventions](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)? No need to write anything too verbose, but I'm very keen to keep the summary line 50 characters only, and separated from the body by two newlines.

ShadowKyogre:
@achadwick: Updated the commit message for the first commit in the patch.

PR #498, opened 2015-11-16
When the user defines a shortcut, always inform Gtk.AccelMap that the…

### LINKED ISSUE #497: Key bindings do not survive a restart unless they're assigned twice.
"Color Details Dialog" and "Edit Frame".

1. Reset all configuration, start MyPaint.
2. Assign a shortcut to one of the aforementioned actions.
3. Restart MyPaint. Shortcuts are no longer there.

It's because MyPaint writes the shortcut as a comment in `accelmap.conf`. This is probably because the relevant action's "changed" field is False.

1. Reset all configuration, start MyPaint.
2. Assign a shortcut to "Colors Details Dialog".  
   MyPaint says `INFO: gui.accelmap: Adding new entry 't': '<Actions>/WindowActions/ColorDetailsDialog'`. [This means "changed" is False](https://github.com/mypaint/mypaint/blob/fd4ea161847f1288f10918bb2c4a6726e594574f/gui/accelmap.py#L488).
3. Assign some other value.  
   MyPaint says `Changing entry 'g': '<Actions>/WindowActions/ColorDetailsDialog'`. [This means "changed" is True](https://github.com/mypaint/mypaint/blob/fd4ea161847f1288f10918bb2c4a6726e594574f/gui/accelmap.py#L482).
4. Restart MyPaint. Shortcut is still here.

So, it's weird that these actions are the only ones not present in `Gtk.AccelMap` by default... but shouldn't it write them in the file anyway?

ydahhrk:
… binding is being overriden, not initialized.

This helps bindings survive restarts regardless of whether the corresponding action is present in the menus or not.
Fixes #497.

I don't know why was this conditional here in the first place, though. Might want to audit before merging.

### DIFF
--- a/gui/accelmap.py
+++ b/gui/accelmap.py
@@ -476,16 +476,11 @@ class AccelMapEditor (Gtk.Grid):
     def _set_accelmap_entry(cls, path, keyval, mods):
         cls._delete_clashing_accelmap_entries(keyval, mods, path)
         accel_name = Gtk.accelerator_name(keyval, mods)
-        entry_exists, junk = Gtk.AccelMap.lookup_entry(path)
-        if entry_exists:
-            logger.info("Changing entry %r: %r", accel_name, path)
-            if Gtk.AccelMap.change_entry(path, keyval, mods, True):
-                logger.debug("Updated %r successfully", path)
-            else:
-                logger.error("Failed to update %r", path)
+        logger.info("Changing entry %r: %r", accel_name, path)
+        if Gtk.AccelMap.change_entry(path, keyval, mods, True):
+            logger.debug("Updated %r successfully", path)
         else:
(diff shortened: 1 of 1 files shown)


achadwick:
Heh, I was just looking at the code myself. Your solution looks neater :laughing: 

Could you possibly make the summary line [shorter than 50 characters](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) and do a forced push to update this PR? Sorry to be picky.

achadwick:
Merged in f28e0a445baccbf186546a4b1e378d8566d2e145, closing #497.

ydahhrk:
Oops, sorry :/
Will do better next time.

achadwick:
No problem, really. Thank you for the fix!

If we have a preferred format for commit messages, we should document it: #499.

Issue #512, opened 2015-11-25
Background chooser dialog: sets default background when it's first opened


loentar:
The problem is background always become white when you re-open saved work with non-white background and select "background chooser" for the fist time after loading. "Undo" does not work at this point and it's not possible to restore background without reverting.

How to reproduce:
1. draw something or leave the picture blank
2. open background chooser and set background
3. save your work
4. restart MyPaint -- required to reproduce
5. open saved work
6. ensure background has a color/texture
7. open background chooser -- background become white.

achadwick:
I cannot reproduce this with a fresh config any MyPaint master as of 0428646094568bb2ed5e034e9e86c5a70c7e2228 (= 1.2.0-beta.3) on Debian testing/unstable in GNOME using GTK+ 3.18.2.

* To reliably get a fresh config, I used

  ```sh
  cd /path/to/clone/of/mypaint
  # ... then rebuild according to the instructions, then...
  rm -vfr /tmp/cfgtmp$$ && ./mypaint -c /tmp/cfgtmp$$
  ```

* When I opened the background chooser, I used the Layer menu.
* When I saved, I used "Save As" (via the Ctrl+Shift+S accelerator)
* MyPaint has no restart action, so I closed the program, waited a short while, then started the program again using

  ```sh
  /mypaint -c /tmp/cfgtmp$$
  ```

* When I opened the file previously saved, I used the Ctrl+O accelerator (File → Open...)
* I am using the default Adwaita theme.

In step 7 of my testing, the background stayed exactly as it was when loaded from the background chooser in step 5.

achadwick:
There is not enough information in your report for me to investigate this bug yet. I need to be able to see it for myself really.

* Please expand on your instructions, if you need to say more.
* Please tell me what
  * operating system + version
  * windowing environment + version
  * version of MyPaint **← most important!** -- be as detailed as you can, always use (Help → About MyPaint... and paste the full version number. If you can, please separate out the hex string on the end - it links up to the revision history nicely on Github if you do that!)
  * version of libgtk **← maybe important**
  * GTK theme
  you are using.
* Please tell me how you are invoking the commands.
* Please tell me if this is still happening for you.

Thanks!

achadwick:
Aha! Needs an extra step zero:

0. Set a default background

and then step 2 needs to be modified:

2. open background chooser and set background <ins>to something other than the default</ins>

So in short, opening the background window sets the background to the default, but only the very first time the background window is opened.

[...]

Issue #499, opened 2015-11-17
Coding conventions: document in one place, add commit message conventions


achadwick:
Our coding conventions are duplicated:

* https://github.com/mypaint/mypaint/blob/dc79f2564ed08e3e36d9e951fc9a5f22efb86582/CONTRIBUTING.md#python-style
* https://github.com/mypaint/mypaint/wiki/Development#getting-contributions-into-mypaint

I suggest merging these docs into the wiki, since coding conventions evolve over time, and are pretty much supplementary information. Note that `CONTRIBUTING.md` is the link given in Github's new submission message:
 
> Please review the [guidelines for contributing](https://github.com/mypaint/mypaint/blob/master/CONTRIBUTING.md) to this repository.

so (IMO) it should remain short and sweet.

----------------------

We should also add some guidelines for commit messages too: <http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html> is a good start.

achadwick:
I've consolidated the coding guidelines and new contributor info onto the wiki, in separate pages for managability. The new index for new contributors is <https://github.com/mypaint/mypaint/wiki/Contributing>.

Still need to update/rewrite <https://github.com/mypaint/mypaint/wiki/Contributing-Patches> to cover commit messages and all the things we want a PR to be.

achadwick:
Updated https://github.com/mypaint/mypaint/wiki/Contributing-Patches, and closing.

PR #515, opened 2015-12-01
Fix issue 512

cortexer:
Possible fix for issue #512:
Simply don't set the background to the window when loading the background chooser dialog.
Normally the background of the paintings is loaded somewhere else (for example in application.py).
Maybe I am overlooking something and this solution is too simplistic?

### DIFF
--- a/gui/backgroundwindow.py
+++ b/gui/backgroundwindow.py
@@ -163,25 +163,13 @@ class BackgroundList (pixbuflist.PixbufList):
         self._background_files.sort()
         self._background_files += self._list_dir(user_path)
 
-        # Load default background
-        self._pixbuf_tooltip = {}
-        defaults = []
+        # Exclude DEFAULT_BACKGROUND from the list shown to the user
         for filename in reversed(self._background_files):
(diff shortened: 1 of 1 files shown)


achadwick:
Hi - there are a couple of issues:

* Can you change the commit message so that the first line is < 50 characters long? It's just a convention and I should really write up what we do but haven't yet (see #499!)
* There seem to be two extra commits which just catch it up to master. These will conflict, but aren't necessary if the change is really just a oneliner. If these aren't part of the fix, can you discard them from this branch please? (Quickhint: you may have to use `git rebase -i`, and do a forced push... this is sort of advanced though.)

I kinda think we need to revisit the entire block though - I'll drop some line comments your way to explain.

[You mentioned](https://github.com/mypaint/mypaint/issues/512#issuecomment-161101218) that this is your first "real" PR, so I can walk you through the stages if you like.

cortexer:
Thanks for the hints.
I improved the PR accordingly.

achadwick:
Looks great - thanks for the fix!

Issue #539, opened 2015-12-28
Cannot run w/o libmypaint-tests.so in /usr/lib?


ShadowKyogre:
Possibly related issues: #535

# Steps to reproduce

1. Package the current git master in a manner similar to `package()` of https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=mypaint-git , sans the copying of libmypaint-tests.so
2. Install the created package (whether it be on Debian, Arch, etc.)
3. Run mypaint.

# Expected result
Runs normally

# Actual result
~~~~
INFO: mypaint: Installation layout: conventional POSIX-like structure with prefix u'/usr'
INFO: mypaint: There is an old-style configuration area in u'/home/shadowkyogre/.mypaint/'
INFO: mypaint: Its contents can be migrated to $XDG_CONFIG_HOME and $XDG_DATA_HOME if you wish.
INFO: mypaint: See the XDG Base Directory Specification for info.
INFO: lib.i18n: POSIX: LANG='en_US'
INFO: lib.i18n: POSIX: LANGUAGE=None
Traceback (most recent call last):
  File "/usr/bin/mypaint", line 462, in <module>
    main.main(datapath, iconspath, old_confpath, version=version)
  File "/usr/share/mypaint/gui/main.py", line 95, in main
    from gui import application
  File "/usr/share/mypaint/gui/application.py", line 48, in <module>
    import lib.document
  File "/usr/share/mypaint/lib/document.py", line 37, in <module>
    import lib.helpers as helpers
  File "/usr/share/mypaint/lib/helpers.py", line 26, in <module>
    import mypaintlib
  File "/usr/share/mypaint/lib/mypaintlib.py", line 28, in <module>
    _mypaintlib = swig_import_helper()
  File "/usr/share/mypaint/lib/mypaintlib.py", line 20, in swig_import_helper
    import _mypaintlib
ImportError: libmypaint-tests.so: cannot open shared object file: No such file or directory

shell returned 1
~~~~

ShadowKyogre:
Is there a way I can get a more in-depth trace of why it's attempting to use libmypaint-tests.so? Especially since when scons sets up an install in a sandbox, it doesn't copy that libmypaint-tests.so automatically.

achadwick:
Patches welcome for getting MyPaint to build and run against sharedlib: see mypaint/libmypaint#28 and more importantly #535 here.

achadwick:
[...]

PR #542, opened 2015-12-30
Check for use_sharedlib before building lib tests

ShadowKyogre:
Addressed #539

Prior to this, it would be impossible to run mypaint without the tests
from brushlib. This shouldn't break the Travis build since it doesn't
set use_sharedlib=true. It surrounds the relevant test code with
ifndef's that check for the presence of NO_TESTS passed from g++ via scons. If
NO_TESTS is defined, then lib/ avoids building any test related code.

It maybe a better idea to define the test building as separate from checking the presence of use_sharedlib as true since there maybe future cases where we may need to run tests with it as a shared lib (instead of assuming shared lib builds are always production builds). 

Let me know if that's the case.

### DIFF
--- a/lib/SConscript
+++ b/lib/SConscript
@@ -20,9 +20,15 @@ def build_py_module(env, *args, **kwargs):
 
 
 # Build against brushlib
-env.Prepend(LIBS=['mypaint-tests', "mypaint"])
+if env['use_sharedlib']:
+    env.Prepend(LIBS=["mypaint"])
+else:
+    env.Prepend(LIBS=['mypaint-tests', "mypaint"])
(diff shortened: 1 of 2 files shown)


achadwick:
Merged as a lightly modified 703fd6b9473f16a8da52e618537c4078e122cc70 to support libmypaint going sharedlib by default (1502bc9e9be16874fa47eda188773f1cbcbeca22 and mypaint/libmypaint#28).

Next step is to ensure that mypaint's remaining tests that need that function from libmypaint-tests get run somehow. This may be quite a task.

Thanks for the code!

Issue #535, opened 2015-12-21
Switch to using libmypaint as a shared lib


achadwick:
MyPaint should build so that it links an installed libmypaint.so at runtime. This will require us to drop the submodule, and a good time to do that is _really early on_ in a new alpha/active dev cycle :sweat_smile: 
### Todo
- [x] Update the docs to say "build libmypaint separately first, and install it" as a first step, essentially
  - MyPaint's C++ parts should use regular shared-lib libmypaint initially
  - I'm not sure about introspection in libmypaint right now. That feels a little incomplete (or maybe I just haven't twiddled the right build vars to get it working properly).
- [x] Do whatever's necessary in the build system to invoke pkg-config to build against and link libmypaint
- [x] Make sure the test runner (`run_brushlib_tests()`) can still be built and run somehow, with the right configuration of libmypaint installed.
- [x] Fix the docs to remove all traces of "brushlib" and the submodule in our build instructions

achadwick:
So libmypaint went sharedlib by default today, which has meant a few changes to mypaint to support it. I note from #539 that we'll need to ensure that the test runner function can be called from MyPaint's CI and testing suite. There's only one function we need, but it's buried in libmypaint-tests.a, which is built but not installed by libmypaint, and which also forms part of its own test suite.

(#542 is a temporary stopgap really. Time to disentangle things properly.)

achadwick:
@Jehan is doing fantastic work making an autotools build in mypaint/libmypaint#32. Disentangled code soon for MyPaint too ☺

Jehan:
@achadwick Please see the PR #538. It has now a new commit which does the "disentangling" into MyPaint! MyPaint does not use any submodule there anymore and build (then run) perfectly. :-)

hartwork:
Hi there!

It would be awesome to have a new release of mypaint that uses external libmypaint. Right now, people can only use _either_ Gimp 2.9.4 (which depends on libmypaint 1.3.x which blocks mypaint <1.2.2) or mypaint (see https://bugs.gentoo.org/show_bug.cgi?id=602066). Please help me unblocking mypaint. Thanks!

Best, Sebastian

odysseywestra:
I think this is technically done @achadwick .

PR #566, opened 2016-01-16
Prepend brushlib directory rather than appending

ryandesign:
This ensures we get brushlib from the source directory, not a different one that might already be installed.

Before this change, MyPaint 1.2.0 builds like this:

    /usr/bin/clang++ -arch x86_64 -arch i386 -o _mypaintlib.so -O3 -dynamiclib lib/mypaintlib_wrap.os lib/fill.os lib/eventhack.os lib/gdkpixbuf2numpy.os lib/pixops.os lib/fastpng.os -L/opt/local/lib -Lbrushlib -L/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib -lmypaint-tests -lmypaint -lm -ljson-c -lpng16 -llcms2 -lgtk-3 -lgdk-3 -lpangocairo-1.0 -lpango-1.0 -latk-1.0 -lcairo-gobject -lcairo -lgdk_pixbuf-2.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0 -lintl -lpython2.7

`-L/opt/local/lib -Lbrushlib` means it will check for libmypaint-tests.a and libmypaint.a in /opt/local/lib first, and only if not found there will it look for them in brushlib. This can cause the build to fail if /opt/local/lib/libmypaint.a already exists, for example from another MyPaint installation. In my case, my already-installed MyPaint was built for x86_64 only, and I was now trying to build MyPaint universal (for x86_64 and i386); the build failed because i386 symbols were not found in the already-installed x86_64 version of the library. But it could fail for other reasons as well.

### DIFF
--- a/lib/SConscript
+++ b/lib/SConscript
@@ -21,8 +21,8 @@ def build_py_module(env, *args, **kwargs):
 
 # Build against brushlib
 env.Prepend(LIBS=['mypaint-tests', "mypaint"])
-env.Append(LIBPATH="../brushlib")
-env.Append(CPPPATH=['../brushlib', '../brushlib/tests'])
+env.Prepend(LIBPATH="../brushlib")
+env.Prepend(CPPPATH=['../brushlib', '../brushlib/tests'])
 
 
 def parse_pkg_config(env, libname):


achadwick:
Okay, that looks sensible for merging onto master. Be aware that we'd like to switch to using libmypaint as a shared lib for 1.3.0 (see #535), but this is useful for now.

Please reformat the comment so that it wraps at 72 characters and generally follows <https://github.com/mypaint/mypaint/wiki/Contributing-Patches#commit-messages>. There's no need to be too fastidious, but I'm fairly picky about consistent formatting ☺

This patch doesn't pull in an updated brushlib submodule. Should it? I can do that as part of the merge, I guess.

A forced push to update the comment would be fine. This PR will update automatically.

achadwick:
See also mypaint/libmypaint#41.

achadwick:
Rebased and merged with the changed mentioned above in 081f624694f9120a65b583f92328e786aa78b69d. Thanks!

ryandesign:
Apologies for not responding; I've been busy at work, and am new to git so I did not know off the top of my head how to do what you asked. Thanks for fixing and merging it.

Issue #595, opened 2016-02-09
Copy and pasting images pastes first copied image


Timmmm:
On Windows:

1. Copy an image (e.g. using the Snipping Tool).
2. Paste it in MyPaint (this works)
3. Copy another image.
4. Paste it in MyPaint

The result is that the first image is pasted. All future pastes paste the first image. You have to quit MyPaint and re-open it to paste a different image.

odysseywestra:
Hi there, thanks for reporting this. Before we can help you, we will need more information. Please visit our [bug reporting wiki](https://github.com/mypaint/mypaint/wiki/Reporting-Bugs) for what information we need from you.

achadwick:
Cannot reproduce with master as of mypaint 5dce1966971ccf2a6832ed33516bdc72cd3653cd in Debian testing/unstable using the GNOME screenshot tool.

@Timmmm Nice steps for reproducing, thanks!
Please let us know your system details, and what version this is when you can.

achadwick:
Here's our paste implementation on the master branch right now: [gui/document.py line 852](https://github.com/mypaint/mypaint/blob/5dce1966971ccf2a6832ed33516bdc72cd3653cd/gui/document.py#L852). There's no variable that stays in scope for long enough for this problem to be cause by caching within MyPaint itself. I think we're handling the [clipboard](https://developer.gnome.org/gtk3/stable/gtk3-Clipboards.html) correctly, but I may be wrong.

(Despite appearances, our code there is PyGI for GTK3)

achadwick:
Note to (potentially just) myself: when testing in a desktop VM, turn off shared clipboards.

Timmmm:
Double checked; this definitely happens. I'm using Windows 10 Pro with 64-bit Mypaint 1.2.0+git.f62444e.dirty.

By the way I'm so surprised Mypaint uses GTK - normally it looks awful on Windows, but Mypaint looks great!

Timmmm:
Looks like this may be an old issue: http://gna.org/bugs/?18278

achadwick:
@Timmmm
Thanks for the info. I don't doubt it's happening. Clipboards are OS/DE-level things, and the GTK backend has to be customised for it. The exact code executed will be different on all systems. I hope it's fixable.

[...]

PR #600, opened 2016-02-13
Clipboard: paste image workaround

cortexer:
When pasting an image from the clipboard always the first image is pasted.
Even if the clipboard has meanwhile been updated.
This commit solves this behaviour.

Closes issue #595


achadwick:
Fixes the problem on Win7. I wonder why? I'll merge for now, and try to investigate what we're supposed to do here.

achadwick:
Merged in 84b807b8c969a3dc96bcaba4085ebfeb794986e1.

PR #633, opened 2016-04-01
transparent pngs: handle default transparency

TuringMachinegun:
Add possibility to save transparent png's by default.
When saving with existent filename (no save dialog), default to saving with transparent background: if the open image has a transparent background, that's the right thing to do. If it doesn't, then the background is opaque anyway (if I am not missing anything).

### DIFF
--- a/gui/filehandling.py
+++ b/gui/filehandling.py
@@ -187,7 +187,8 @@ class FileHandler (object):
         self.config2saveformat = {
             'openraster': SAVE_FORMAT_ORA,
             'jpeg-90%': SAVE_FORMAT_JPEG,
-            'png-solid': SAVE_FORMAT_PNGSOLID,
+            'png-trans': SAVE_FORMAT_PNGTRANS,
+            'png-solid': SAVE_FORMAT_PNGSOLID
         }
 
         self.__statusbar_context_id = None
@@ -798,7 +799,7 @@ class FileHandler (object):
         if not self.filename:
             self.save_as_cb(action)
         else:
-            self.save_file(self.filename)
+            self.save_file(self.filename, alpha=True)  # defaults to transparent background
 
     def save_as_cb(self, action):
         if self.filename:
@@ -863,10 +864,15 @@ class FileHandler (object):
                     else:
                         saveformat = default_saveformat
 
-                # if saveformat isn't a key, it must be SAVE_FORMAT_PNGAUTO.
-                desc, ext_format, options = self.saveformats.get(saveformat,
-                    ("", ext, {'alpha': None}))
-                #
+                if saveformat == SAVE_FORMAT_PNGAUTO:
+                    if self.app.preferences['saving.default_format'] == 'png-trans':
+                        alpha = True
+                    else:
+                        alpha = None  # whether it will end up to be true or false depends on visibility of background
+                    desc, ext_format, options = ("", ext, {'alpha': alpha})
+                else:
+                    desc, ext_format, options = self.saveformats[saveformat]
+
                 if ext:
                     if ext_format != ext:
                         # Minor ugliness: if the user types '.png' but
--- a/gui/preferenceswindow.glade
+++ b/gui/preferenceswindow.glade
@@ -24,6 +24,10 @@
         <col id="1" translatable="yes" context="Prefs Dialog|Saving|Save Formats and Locations|Default|">OpenRaster (*.ora)</col>
       </row>
       <row>
+        <col id="0">png-trans</col>
+        <col id="1" translatable="yes" context="Prefs Dialog|Saving|Save Formats and Locations|Default|">PNG transparent (*.png)</col>
+      </row>
+      <row>
         <col id="0">png-solid</col>
         <col id="1" translatable="yes" context="Prefs Dialog|Saving|Save Formats and Locations|Default|">PNG solid with background (*.png)</col>
       </row>


odysseywestra:
Thanks for the pull request! Well get this tested and let you know if we need you make any changes. Question which issue(s) did this address?

TuringMachinegun:
Great, thanks! First contribution to an open source project for me :)

I don't think it solves any issue in the issue tracker. It is adding a feature I needed, that is the possibility to save transparent pngs by default, and not only the solid ones. I guess other people may want that. My use case is that I use mypaint to make diagrams which I then import in emacs org-mode, and it is visually much more appealing to have the transparent background. It was an hassle for me to select the possibility manually every time.

One thing I changed which I think could be problematic is that I made the background transparent by default when an image is saved through "save" with no dialog. This is because if I wanted to modify my handwritten notes and then save them, an opaque background would appear. My reasoning is that if the figure already had a transparent background, then all is fine; if it didn't, the background will be opaque anyway.
But I'm not completely sure about this last point, I'm not sure whether there are two types of png formats, with and without transparency, and mypaint wants to keep them separate. In that case, saving with default transparency maybe unwanted behaviour, and mypaint should check if the open png is of the transparent type or not.
In case this is unwanted behaviour, it is only a line that needs modifying.

gui/filehandling.py
```diff
@@ -798,7 +799,7 @@ def save_cb(self, action):
         if not self.filename:
             self.save_as_cb(action)
         else:
-            self.save_file(self.filename)
+            self.save_file(self.filename, alpha=True)  # defaults to transparent background
```
odysseywestra: I'm getting an error that probably caused by this.

Steps to Reproduce:
1) Open with a Blank Canvas, and scribble on it
2) Save as Ora
3) Save again with File>Save

**Traceback Dump:**
```python
Mypaint version: 1.3.0-alpha+git.9f0e6f8
System information: Linux-4.2.0-34-generic-x86_64-with-Ubuntu-15.10-wily
Using: Python 2.7.10, GTK 3.16.7, GdkPixbuf 2.32.1, Cairo 1.14.4
  File "/home/albert/Public/Github/mypaint/gui/filehandling.py", line 802, save_cb(self=<gui.filehandling.FileHandler object>, action=<Action object at 0x7f90af588320 (GtkAction at 0x1aa51f0)>)
            else:
                self.save_file(self.filename, alpha=True)  # defaults to transparent background
  variables: {'alpha': (None, []), 'self.save_file': ('local', <bound method FileHandler.save_file of <gui.filehandling.FileHandler object at 0x7f90af57a550>>), 'True': ('builtin', True), 'self.filename': ('local', u'/home/albert/Artwork/Current Projects/Linux Showdown/Source Files/2016-0330-Team_Noah02.ora')}
  File "/home/albert/Public/Github/mypaint/gui/drawwindow.py", line 85, wrapper(self=<gui.filehandling.FileHandler object>, *args=(u'/home/albert/Artwork/Current Projects/Linux Showdown/Source Files/2016-0330-Team_Noah02.ora',), **kwargs={'alpha': True})
            try:
                func(self, *args, **kwargs)
                # gtk main loop may be called in here...
  variables: {'self': ('local', <gui.filehandling.FileHandler object at 0x7f90af57a550>), 'args': ('local', (u'/home/albert/Artwork/Current Projects/Linux Showdown/Source Files/2016-0330-Team_Noah02.ora',)), 'func': ('local', <function save_file at 0x7f90af60a7d0>), 'kwargs': ('local', {'alpha': True})}
  File "/home/albert/Public/Github/mypaint/gui/filehandling.py", line 555, save_file(self=<gui.filehandling.FileHandler object>, filename=u'/home/albert/Artwork/Current Projects/Linux Showdown/Source Files/2016-0330-Team_Noah02.ora', export=False, **options={'alpha': True})
                statusmsg=True,
                **options
            )
  variables: {'options': ('local', {'alpha': True})}
  File "/home/albert/Public/Github/mypaint/gui/filehandling.py", line 636, _save_doc_to_file(self=<gui.filehandling.FileHandler object>, filename=u'/home/albert/Artwork/Current Projects/Linux Showdown/Source Files/2016-0330-Team_Noah02.ora', doc=<gui.document.Document object>, export=False, statusmsg=True, **options={'alpha': True, 'save_srgb_chunks': True})
                    feedback_cb=self.gtk_main_tick,
                    **options
                )
  variables: {'options': ('local', {'alpha': True, 'save_srgb_chunks': True})}
  File "/home/albert/Public/Github/mypaint/lib/document.py", line 1212, save(self=<Document nlayers=32 bbox=Rect(0, 0, 2176, 3456) paintonly=False>, filename=u'/home/albert/Artwork/Current Projects/Linux Showdown/Source Files/2016-0330-Team_Noah02.ora', **kwargs={'alpha': True, 'feedback_cb': <function gtk_main_tick>, 'save_srgb_chunks': True})
            try:
                result = save(filename, **kwargs)
            except GObject.GError as e:
  variables: {'kwargs': ('local', {'feedback_cb': <function gtk_main_tick at 0x7f90af60a5f0>, 'alpha': True, 'save_srgb_chunks': True}), 'save': ('local', <bound method Document.save_ora of <Document nlayers=32 bbox=Rect(0, 0, 2176, 3456) paintonly=False>>), 'result': ('local', None), 'filename': ('local', u'/home/albert/Artwork/Current Projects/Linux Showdown/Source Files/2016-0330-Team_Noah02.ora')}
  File "/home/albert/Public/Github/mypaint/lib/fileutils.py", line 118, _wrapped_save_method(self=<Document nlayers=32 bbox=Rect(0, 0, 2176, 3456) paintonly=False>, filename=u'/home/albert/Artwork/Current Projects/Linux Showdown/Source Files/2016-0330-Team_Noah02.ora', *args=(), **kwds={'alpha': True, 'feedback_cb': <function gtk_main_tick>, 'save_srgb_chunks': True})
                    logger.error("cleanup: failed to remove temp path too")
                raise ex
            if not os.path.exists(temp_path):
  variables: {'ex': ('local', TypeError("save_as_png() got multiple values for keyword argument 'alpha'",))}
TypeError: save_as_png() got multiple values for keyword argument 'alpha'
Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.
```
I did remove the change you made in that line and that removed the error. However MyPaint freezes up randomly when I try to save.

TuringMachinegun:
Ooops. Sorry about that. I should have tested it more carefully. Yeah, that line is not going to be of use. I wonder what's happening with the random freezes though. I'll look into that

odysseywestra:
Okay. Let me know when you are ready for me to test it out again. ^^

TuringMachinegun:
Hey, I couldn't reproduce the freezes, can you give me some more details about the conditions in which they happen?

odysseywestra:
Let me re-test everything since my lag issue may be related to #644.

odysseywestra:
Okay I couldn't reproduce the freezes either with a single monitor, So I don't think it's related to this PR.

gui/filehandling.py
odysseywestra: Found out that the random freezing is related to #644 and not this PR.

odysseywestra:
@TuringMachinegun just checking in to see how are coming around with this PR.

TuringMachinegun:
Hey! Well, I'm not sure what I should do next. If the freezes were related to #644 then I suppose the code I submitted works fine? It does for me. The offending alpha=True should still be left out because it's just wrong (as I didn't account for .ora's, quite the overlook...!), but as far as I can see the rest of the code shouldn't be cause of headaches.
I was thinking that to solve the problem that I was trying to solve with alpha=True when opening existing png's (that is, automatically leave the background transparent if found transparent), one needs to check whether an alpha channel is present in the .png. Which can be done quite easily in Python as far as I can see, but I'm not sure I know enough of mypaint inner workings to implement it in the right way. Also, I'm not sure whether it is a desired feature or not, there might be reasons I don't see why it wouldn't be the mypaint way of doing things.

odysseywestra:
The only problem with removing that segment, is when you save it again as PNG, it will not save the transparency even with PNG transparent as Default saving. The opposite happens you that is added, it will save the transparency even though the solid background is selected as default. 

@achadwick could you elaborate on how the `Default File Format` option in edit is suppose work?

achadwick:
No problem!

The `saving.default_format` config value is only used when saving normally when you're letting MyPaint choose based on the file name extension you give the file. Or if the file has no extension.

The potential values are hardcoded in `default_save_format_liststore` in `gui/preferenceswindow.glade`. They have to be keys in the main FileHandler object's `self.config2saveformat` dict.

This configuration option is also used when saving scrap files and scratchpad files, so take care! This is an area that could do with some consolidation and more careful thought anyway (who the heck wants PNG files in those areas?) but let's hold off on that for now...

TuringMachinegun:
Alright! The code I submitted doesn't deal with scrap files and scratchpad files, so I'll have to go through it again.
@achadwick, what about checking for transparency when opening png files, so that mypaint automatically knows what to do when saving an existing png that was open for editing? Is that something somebody is already working on / is there a demand for that / does it make sense to have it?

odysseywestra:
Another way you could indicate transparency as default is to see if the background layer is on or off. When I save an image that has the background layer off, when I save it as PNG, I would expect to have a transparent background. MyPaint doesn't do that at the moment so you could do it that way as well.

odysseywestra:
@TuringMachinegun You can close this if you think this needs more work. If you have any questions or need help, feel free to open a thread on our Community Forums under the Developers Area.

http://community.mypaint.org/c/development

TuringMachinegun:
Ok great, I'll work on this some more. Oh nice, I didn't know about the forum, thanks!

Issue #298, opened 2015-04-21
Sidebar panels (docked subwindows) don't remember their sizes after resizing the main window


Griatch:
1. Have a brush list (maybe other docks, haven't tested) open in a toolbar and extend it vertially as desired.
2. Minimize MyPaint, then maximize it back up again
3. The brush list is now shrunk back to its smallest possible vertical size. One has to re-size it manually. 

It feels more reasonable that dock proportions should be retained across minimization. 

Tested with latest MyPaint GIT version on Linux Mint (KDE).

achadwick:
Specific MyPaint version and OS version please?

Good suggestion though. Bugs me too.

achadwick:
Likely to not have string changes. Beta cycle, maybe.

Griatch:
Specific OS version is Linux Mint 17 with KDE desktop.

achadwick:
I think I may be misinterpreting this report and thinking of something else.

* When you say a "toolbar", do you mean a floating window, or a docked sidebar?
* When you say "minimize", what do you mean? Which button or menu item are you selecting?
* What specific version of MyPaint is this? Please look in the about box and report the string there.

I cannot reproduce this one with git master at 0b3067fcb29312135aa14d8e90076611cc2b3f01 (Tue Apr 21), in GNOME 3.14, trying "Minimize" on the main window from the menu (yay, GNOME doesn't have a minimize button)

![sshot](https://cloud.githubusercontent.com/assets/61299/7257381/2fb4d3f4-e84d-11e4-83f6-5fbad7958f12.png)

Also:

* Is this a very recent change? I need to know whether it works with the previous commit, c0c088b7391ecb9c70c1d21ab4d81b3bd0b0884d, then fails with 0b3067fcb29312135aa14d8e90076611cc2b3f01 because window management changed with that commit.

achadwick:
Cannot reproduce in Linux Mint KDE 17.1 LTS as described in the original report, under a virtualbox VM and with all the default Linux Mint KDE Edition stuff. Here's what I did:

1. Opened one brush list in the sidebar, and one in a floating window, a bit like the GNOME screenshot above.
2. Resized both brush list windows down a bit, to show all brushes
3. Clicked on the main window's minimize button:
    * ![kwin minimize button](https://cloud.githubusercontent.com/assets/61299/7265604/2140d42a-e894-11e4-8faa-74ad737b9561.png)
    * Application is minimized to the panel
4. Clicked on the button for the window in the panel
[...]

PR #650, opened 2016-04-15
ToolStack: Fix sticking of the last paned.

danpla:
The problem is that when a widget is created, its parent Gtk.Container is already connected to the "size-allocate" signal of the parent Gtk.Widget. Excerpt from https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-size-allocate:

> This function is only used by GtkContainer subclasses, to assign a size and position to their child widgets.

The simple solution is to override do_size_allocate() so that our handler will be called first.
Another good thing is that the separator is now sticks when it's moved manually.

Fixes mypaint/mypaint#298.


achadwick:
Looks quite an elegant fix. Thank you!

PR #668, opened 2016-04-30
Replace Gtk.Widget.drag_source_set_icon_pixbuf() with Gtk.drag_set_icon_pixbuf()

danpla:
Gtk.drag_set_icon_pixbuf() is the right way to set the D&D icon
dynamically from the "drag-begin" event. drag_source_set_icon_pixbuf()
should be used to set a "static" icon.

In theory, they should work equally, but I noticed that icon set via
drag_source_set_icon_pixbuf() is not visible if its size is greater
than 55px (as in the brush list). Maybe it's a bug in GTK+ or
something desktop-specific (XFCE), but anyway it's better to use
the "right" way.

Examples from the GTK+ sources:
[gtk_drag_set_icon_pixbuf](https://github.com/GNOME/gtk/search?utf8=%E2%9C%93&q=gtk_drag_set_icon_pixbuf&type=Code)
[gtk_drag_source_set_icon_pixbuf](https://github.com/GNOME/gtk/search?utf8=%E2%9C%93&q=gtk_drag_source_set_icon_pixbuf&type=Code)

### DIFF
--- a/gui/brushselectionwindow.py
+++ b/gui/brushselectionwindow.py
@@ -125,7 +125,7 @@ class BrushList (pixbuflist.PixbufList):
         preview = preview.scale_simple(
             preview.get_width()//2, preview.get_height()//2,
             GdkPixbuf.InterpType.BILINEAR)
-        self.drag_source_set_icon_pixbuf(preview)
+        Gtk.drag_set_icon_pixbuf(context, preview, 0, 0)
         super(BrushList, self).drag_begin_cb(widget, context)
 
     def on_drag_data(self, copy, source_widget, brush_dragid, target_idx):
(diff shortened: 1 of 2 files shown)


achadwick:
For me (GNOME, Debian testing/unstable, GTK at 3.18.9; MSYS2, Windows 7, GTK at 3.20.3) it's the other way round with master at 8bebf27a031f49538f62832df7044e5d5015ec7a. Brushes (big pixbufs) have a correct drag icon, and colours (small pixbufs) do not.

But this PR fixes that quite nicely, and also suppresses the assertions below. Thank you!

```
(mypaint:23911): Gtk-CRITICAL **: gtk_drag_source_set_icon_pixbuf: assertion 'site != NULL' failed
(mypaint:23911): Gtk-CRITICAL **: gtk_drag_source_set_icon_pixbuf: assertion 'site != NULL' failed
```

With or without this patch, the `win32` backend for 3.20.3 draws the drag icon at the display's (0, 0) at all times during the drag, making the drag hard to see happening. However with your fix, both icons are now correct (despite still being in the wrong place).

PR #672, opened 2016-05-04
BackgroundList: Fix preview scale factor.

danpla:
In Python 2, int(x) / int(y) always returns 0 if y > x.


achadwick:
I don't see much difference, but I guess this is safe... ☺

pull672

danpla:
Try with 128x128 background (upscaled `03_check2.png`):
03_check3

Before/after:
2016-05-14-225451_1600x900_scrot 2016-05-14-225541_1600x900_scrot

achadwick:
Thanks for the explanation too :smile:

PR #688, opened 2016-05-30
Standardize NumPy imports

QuLogic:
The convention is `import numpy as np`, so use that instead of the various other forms.

### DIFF
--- a/gui/freehand.py
+++ b/gui/freehand.py
@@ -11,8 +11,6 @@
 ## Imports
 
 import math
-from numpy import array
-from numpy import isfinite
 from lib.helpers import clamp
 import logging
 from collections import deque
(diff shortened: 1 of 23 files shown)


gui/freehand.py
```diff
@@ -772,8 +772,8 @@ def _interpolate_p0_p1(self):
                 t, x, y = np[0:3]
                 p, xt, yt = spline_4p(
                     float(t - t0) / dt,
-                    array(pt0p[3:]), array(pt0[3:]),
-                    array(pt1[3:]), array(pt1n[3:])
+                    np.array(pt0p[3:]), np.array(pt0[3:]),
+                    np.array(pt1[3:]), np.array(pt1n[3:])
```
achadwick: I can't merge this PR yet due to the variable shadowing here.

achadwick:
@QuLogic 
Please can you rebase against master then make sure the test case still runs? Thanks.

gui/freehand.py
QuLogic: Oops, should be fixed now.

achadwick:
Thanks, looking good. :+1:

PR #689, opened 2016-05-30
Python 3, stage 1 compatibility

QuLogic:
Most of stage 1 from [`futurize`](http://python-future.org/). I did not apply the `absolute_import` setting yet because mypaint is rife with relative imports and that would balloon this PR. Also, haven't fully reviewed the `division` setting yet.

These changes are confined to the files in which they actually make a difference. I will put blanket changes like enabling `print_function` across the code base in a separate PR where it will be easier to see just those changes.

This depends on #688 due to conflicts.

### DIFF
--- a/gui/application.py
+++ b/gui/application.py
@@ -444,7 +444,7 @@ class Application (object):
             jsonstr = open(settingspath).read()
             try:
                 return helpers.json_loads(jsonstr)
-            except Exception, e:
+            except Exception as e:
                 logger.warning("settings.json: %s", str(e))
                 logger.warning("Failed to load settings: using defaults")
                 return {}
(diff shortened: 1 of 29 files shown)


QuLogic:
Rebased to fix conflicts.

achadwick:
Okay, thanks again. I did a quick check to make sure everything's OK, hope I've not missed anything.

KuyaMarc:
I made a Quick Fix...
I don't know if someone already has a solution, but I quickly made a "Quick Fix".

When I see the following:
```
  File "/usr/local/bin/mypaint", line 29
    from __future__ import print_function
SyntaxError: from __future__ imports must occur at the beginning of the file
```

I enter the following statement:
**`sudo sed -i -e 's/from __future__/#from __future__/g' /usr/local/bin/mypaint`**

Then, of course, I am using Debian Linux and am not printing!

I am compiling & testing MyPaint in the following systems, with my Genius EasyPen i405X.
- 64-bit Debian Linux 9 "Stretch" 
- 32-bit Raspbian Linux 8 "Jessie" (Raspberry Pi Model B 512MB)
- 32-bit Debian Linux 8 "Jessie"

And I don't have a printer. Have a Nice Day!

PR #690, opened 2016-05-30
 	py3k: Add print __future__ import to all files.

QuLogic:
This follows on from #689, so it's pretty large; it'll make more sense once that one is merged. Only the last commit is really important here.


achadwick:
Thanks again ☺

PR #691, opened 2016-05-30
py3k: Use new-style division

QuLogic:
On Python 3, division returns floats, while `//` returns integers. In cases where there are explicit casts to `float`, or `float` literals, I have left the plain `/`. In cases where there are explicit casts to `int`, I've changed it to `//`.

I also assumed that bboxes, widths, and heights are all integers and used `//`, but if that's not the case, I can re-check those cases.

I'll add a few comments where I'm not sure.

### DIFF
--- a/gui/colors/adjbases.py
+++ b/gui/colors/adjbases.py
@@ -10,7 +10,7 @@
 """Manager+adjuster bases for tweaking a single color via many widgets.
 """
 
-from __future__ import print_function
+from __future__ import division, print_function
 
 import math
 from copy import deepcopy, copy
@@ -834,7 +834,7 @@ class IconRenderableColorAdjusterWidget (ColorAdjusterWidget, IconRenderable):
         suggested small outer border.
 
         """
-        b = max(2, int(size / 16))
+        b = max(2, int(size // 16))
         self.render_background_cb(cr, wd=size, ht=size, icon_border=b)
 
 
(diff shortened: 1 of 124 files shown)


gui/colors/adjbases.py
```diff
@@ -876,10 +876,10 @@ def render_background_cb(self, cr, wd, ht):
         cr.set_line_width(self.OUTLINE_WIDTH)
         cr.stroke()
 
-        cr.rectangle(b, b, int(eff_wd / 2), eff_ht)
+        cr.rectangle(b, b, int(eff_wd // 2), eff_ht)
         cr.set_source_rgb(*curr.get_rgb())
         cr.fill()
-        cr.rectangle(wd / 2, b, eff_wd - int(eff_wd / 2), eff_ht)
+        cr.rectangle(wd // 2, b, eff_wd - int(eff_wd // 2), eff_ht)
```
QuLogic: `cr.rectangle` can take a float, so I'm not sure if it should be an integer or not.

achadwick: Our Cairo drawing often uses 0.5px offsets to make 1px lines align to the pixel grid. The worst that can result is a fuzzy line though.

QuLogic: So should I put it back to float division, then? I'm not sure whether `wd` is or is not a float at this point.

odysseywestra:
@QuLogic could you rebase this and resolve the conflicts on this one before we merge it in?

QuLogic:
Rebased.

odysseywestra:
@QuLogic  Thank you.
@achadwick Anything you want to address before merging this in? Like the comment you had in `gui/colors/adjbases.py `?

achadwick:
Thanks for the rebase, and I'm sorry this has taken so long. Pushed.

I've been testing it out, and the colour selectors all look nice and sharp. No exceptions thrown so far :smile:

Issue #10, opened 2014-06-23
Background alignment: appears to shift when changed after reloading an OpenRaster file


achadwick:
Migrated here from [Gna! bug 16741](https://gna.org/bugs/?16741), where it was originally reported as a transparent PNG issue (sorry, WONTFIX for transparent PNGs: they don't have any background expectations and shouldn't).

However it evolved into a bug which can be provoked when using the layered OpenRaster format, which is intended to alleviate this sort of problem. Steps to reproduce:

1. Pick a "comic panel" type background
2. Paint in a panel, more than 64 pixels from the top and right of the "panel"
3. Save as .ora
4. Load in the same .ora
5. Open the background selection window again, choose another background
6. Choose the  "comic panel" type background again
  - background has now shifted irrevocably

There's no need to close the background selection window between steps 5 and 6: the bug presents immediately.

Not sure how much we want to chase this for 1.2 - it's obviously caused by the save code only saving populated tiles (and the background underneath them). The save alignment stuff frankly scares me though, particularly when the frame is factored in; I'd rather have weird behaviour when reassigning backgrounds than subtle alignment bugs emerging between saving and loading.

achadwick:
I've tickled the saving alignment since I migrated this one over, but this is still happening in as of 18b7287fd6739b34015808c90bfbfdc7d29bf511.

---------------------

One workaround for users bitten by this bug right now is to create a layer group at the bottom of their document, move each layer one by one into it, and then move the group to the corrected position.

---------------------

Possible long-term fix would be to expose the background as an ordinary layer and allow it to be moved.

---------------------

**Triage note**: not stale, still happening, has workaround, has reliable steps to reproduce.

PR #693, opened 2016-05-31
Use the right bounding box when saving a file

danpla:
Currently, unless the frame is enabled, MyPaint saves files using an internal (dynamic) bounding box, which is always multiple by the tile size, and has (0, 0) coordinates. This is not the right way to do this. Let me explain using ORA as an example.

There are two backgrounds that MyPaint writes to ORA files:

* External, saved as a regular layer to be used by other software. This layer is just cropped using dimensions and coordinates of the dynamic bounding box. This can be OK when the background has no sensible borders (like a seamless canvas texture), but otherwise this is not what the user expects.

* Internal, to be used by MyPaint. This one has the right size (the size of the original image used as background), but since it's cropped from coordinates of the dynamic bounding box, it becomes completely broken (see issue #10).

From the user's point of view, in both cases MyPaint writes a corrupted ORA, because it's not possible to restore the previously used background image, neither in MyPaint (see #10) nor in any other software.

The user should not care about MyPaint internals (the tile size in this case).

For example, when I use `50_comicthumbs.png` for background, I expect that MyPaint will save the full frame, even if I drew something in the bottom right corner (or didn't draw anything at all). In other words, the final size of the saved image should always be multiple of the size of the background image.

### DIFF
--- a/lib/document.py
+++ b/lib/document.py
@@ -1035,6 +1035,60 @@ class Document (object):
         """
         return self.get_frame() if self.frame_enabled else self.get_bbox()
 
+    def get_user_bbox(self):
+        """Return the bounding box expected by the user.
+
+        If the frame is enabled, this is the bounding box of the frame.
+
+        If the frame is disabled, this is a rectangle multiple of the size
+        of the current background image, big enough to cover the dynamic
+        bounding box of the document. Width and height will always be
+        greater than zero, that is, the minimum size is the size of the
+        background image.
+        """
+        if self.frame_enabled:
+            return self.get_frame()
+
+        bbox = self.get_bbox()
+        bg_bbox = self.layer_stack.background_layer.get_bbox()
+        assert bg_bbox.w > 0
+        assert bg_bbox.h > 0
+
+        # Note: In Python, -11 % 10 == 9
+
+        correction_x = bbox.x % bg_bbox.w
+        if correction_x > 0:
+            bbox.x -= correction_x
+            bbox.w += correction_x
+
+        correction_y = bbox.y % bg_bbox.h
+        if correction_y > 0:
+            bbox.y -= correction_y
+            bbox.h += correction_y
+
+        if bbox.w == 0:
+            bbox.w = bg_bbox.w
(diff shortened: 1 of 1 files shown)


achadwick:
Liking this so far. It also makes Nautilus thumbnails for comics more understandable :+1: 

screenshot from 2016-06-08 00-45-13

achadwick:
I would be testing with GIMP, but its ORA plugin is currently too broken to load *any* .ora file on my test machine. Bah. The offsets in `stack.xml` look Perfectly Reasonable to me at a first glance :smiley: 

Testing: I need to confirm that:

- [ ] Files saved with `user-bbox`still open and looking correct under the current master as of 8acacb3dd0d972371d8b0a97176e5a42d011b25f.
- [ ] Hand-hacked versions of a file saved with _user-bbox_ *without* a src attr on the background layer load OK in 
  - [ ] master as of 8acacb3dd0d972371d8b0a97176e5a42d011b25f
  - [ ] user-bbox
- [ ] Hand-hacked versions of a file saved with _user-bbox_ *relying only on* the src attr on the background layer load OK in
  - [ ] master as of 8acacb3dd0d972371d8b0a97176e5a42d011b25f
  - [ ] user-bbox

achadwick:
All looks good. Thank you for closing this long-standing bug! :grinning:

0ion9:
@achadwick: too broken to load any ora, really? I've been working with a version of file-openraster.py that simply has one line added to load_ora(): `img.layers[-1].set_offsets(0,0)` (just before `return img`). That's been working quite well, I guess I'll have to remove it now.

danpla:
@0ion9 On my machine:
```python
  File "/usr/lib/gimp/2.0/python/gimpfu.py", line 851, in _run
    res = apply(func, params[1:])
  File "/usr/lib/gimp/2.0/plug-ins/file-openraster.py", line 274, in load_ora
    if not name:
UnboundLocalError: local variable 'name' referenced before assignment
```
This has been fixed in the GIMP's master branch.

PR #694, opened 2016-05-31
Pan shortcuts: allow user to customize keys.

ydahhrk:
Pan Left, Pan Right, Pan Down and Pan Up used to be hardcoded into the arrow keys. I find this really awkward, as I need to pan often, my shortcut hand is sitting in the opposite side of the keyboard, and the spacebar option is a three step process my brain refuses to get used to. (Hit space, pan, hit space again.)

This commit allows the user to customize the four directional pan actions.

Objections?
Please review this change in particular:
https://github.com/ydahhrk/mypaint/commit/a165e6ab1275da0d4debf507ec2d235803ef8514#diff-43086ad29962d3495933ef1c3bc3ad5bL400

### DIFF
--- a/gui/accelmap.py
+++ b/gui/accelmap.py
@@ -397,8 +397,10 @@ class AccelMapEditor (Gtk.Grid):
         # So we get (<Shift>j, Shift+J) but just (plus, +). As I
         # understand it.
 
-        if not Gtk.accelerator_valid(keyval, mods):
-            return True
+        # This is rejecting some legit key combinations such as the
+        # arrowkeys, so I had to remove it...
+        #if not Gtk.accelerator_valid(keyval, mods):
+        #    return True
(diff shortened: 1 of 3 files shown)


achadwick:
I don't mind adding extra actions that the user can change the binding for. Even when their default binding is Left/Right/Up/Down.

But please leave the existing "extra" keymap bindings for the cursor keys as they are. It's like a fallback layer for when those keys aren't bound in the "main" keymap. It'll allow the cursor keys to be used even if WASD or whatever is bound in the main keymap.

ydahhrk:
k

ydahhrk:
Ok, done

achadwick:
Looks good to me, squashed and merged. I hope the commented-out test doesn't break anything.

PR #697, opened 2016-06-01
RootLayerStack.render_thumbnail: Ensure that the smaller side is > 0.

danpla:
With a frame like 10x1000px, the smaller scaled side becomes zero, and `pixbufsurface.Surface` raises an assertion error.


achadwick:
Good spot, thanks!

PR #701, opened 2016-06-03
Layer groups: Propagate properties to children.

### LINKED ISSUE #699: Groups don't propagate properties to children
You can make a group invisible or lock it, but it will still be possible to change its children: paint, move, fill etc.

It should be possible for any item in a stack to somehow access the topmost group (but *not* the root stack itself) to return the right values from `visible`, `locked` and other properties. This also means that a group should *not* change properties of its children.

Krita can be used as a reference.

danpla:
I leave the FloodFill mode as is. Its behavior is especially buggy: it always creates a new layer on top of the locked, even if it's in a locked branch. If this is an acceptable PR, should I also add a commit to allow using the FloodFill only on paintable layers, even if "New Layer" is enabled? Or it's better to create a separate ticket on the issue tracker?

Fixes #699.


achadwick:
Very neat so far, and it seems to work out fine despite my misgivings about having a group reference for layers.

Would it be possible for you to split this commit into more minimal, working commits? Ideally, one commit should do one thing and one thing only.

lib/layer/core.py
```diff
@@ -237,6 +238,20 @@ def clear(self):
     ## Properties
 
     @property
+    def group(self):
```
achadwick: Please add some doctests (like `root`'s) for this property. In particular, the doctest _is_ documentation, so it should readably test which group of a nested structure a layer's `group` property refers to.

image

We need to be pretty fanatical about CI checks now the project is growing.

gui/freehand.py
```diff
@@ -657,9 +657,8 @@ def _process_queued_event(self, tdw, event_data):
             else:
                 drawstate.avgtime = (tavg, nevents)
 
-        # Refuse drawing if the layer is locked or hidden
         current_layer = model._layers.current
-        if current_layer.locked or not current_layer.visible:
+        if not current_layer.get_paintable():
```
achadwick: This recursive test might too much to do on every event, even in queues.

Also (and this is a more important shortcoming of the original code), the queued event handler should not be looking up the model's current layer on each queued event. Or even the target layer. What if the user has changed layers while there is a big queue to process? I think things are meant to flush before that happens, but it seems cleaner to avoid the *possiblity* of error too.

(So that's two problems... feel up for a fix of my code too?) :grinning: 

IMO, modes should observe the model more than they do right now. Painting modes should respond to potential changes of the get_paintable() state of the current layer by updating a boolean property. Use the boolean property at capture only, if capture is separate from painting. And FreehandMode should probably cache a target layer for its painting too, if that's possible.

lib/layer/data.py
```diff
@@ -1322,6 +1302,22 @@ def _load_strokemap_from_ora(self, elem, x, y, orazip=None, oradir=None):
...
+        """True if this layer currently accepts painting brushstrokes"""
+        return (
+            self.visible and
+            not self.locked and
+            self.branch_visible and
+            not self.branch_locked
+        )
+
+    def get_fillable(self):
+        """True if this layer currently accepts flood fill"""
```
achadwick: Would it be possible to add some doctests for these?

Not sure if it's an error or not, but why allow filling on functionally invisible layers?

gui/layers.py
```diff
@@ -754,41 +750,38 @@ def layer_visible_pixbuf_datafunc(column, cell, model, it, data):
...
+            sensitive = False
+        else:
+            visible = layer.visible
+            sensitive = layer.branch_visible
+
+    icon_name = "mypaint-object-{}-symbolic".format(
+        "visible" if visible else "hidden",
     )
     cell.set_property("icon-name", icon_name)
+    cell.set_property("sensitive", sensitive)
```
achadwick: Thanks. The name hack always seemed an ungreppable mess to me, and this change doesn't seem to have any bad consequences like making the icon unclickable. At least it doesn't for Adwaita and GTK 3.20.4 :eyes:

danpla:
@achadwick  Yes, I'll split it.

lib/layer/data.py
danpla: As I mentioned, I leave FloodFill as is (well, except this method - I'll remove this). I think fixing it is out of the scope of this PR. Such a fix will change its current behavior quite noticeable, and this requires some discussion. I think it's better to create a separate issue/PR.

gui/freehand.py
danpla: It looks like this line is redundant, since it's not possible to queue an event when the current layer is not paintable.

Yes, currently, if you will use a brush with gigantic size, it's possible to change layers while FreehandMode process the queue. In this case, parts of a stroke will be on separate layers. But I think that such a scenario is uncommon, so it's not critical issue (for example, compared to the #699). Anyway, the `BrushworkModeMixin` is responsible for creating brush strokes, and it's always place them on the current layer. Fixing this requires quite big changes, which is out of scope of this PR — we need a separate ticket on the issue tracker.

danpla:
Done.

PR #714, opened 2016-06-27
Fix coordinates of the "object-unlocked" icon

danpla:
Before/after:
img


achadwick:
I'd like to see what happens if we keep the body of the lock still, but reflect the shackle popping out.

danpla:
Like in Krita?
2016-06-27-154115_1600x900_scrot

odysseywestra:
This is a fairly simple fix. @achadwick are you okay with me Rebasing and Merging this one in?

achadwick:
Yes, fine. Let's do that now, in fact. I've been putting this off because I quite like the body of the lock popping out like that, in a way.

We can always revert later on :)

Issue #712, opened 2016-06-27
Layers: Double click on expanders emits renaming


danpla:
#### Description of the problem

A double click on expander emits renaming of a layer, which is pretty annoying. This happens because expanders attached to "name" columns, which request rename on a double click.

There is a simple workaround using [Gtk.TreeView.get_cell_area()](http://lazka.github.io/pgi-docs/Gtk-3.0/classes/TreeView.html#Gtk.TreeView.get_cell_area), after which renaming will only happen after double click on the areas highlighted in green:
![2016-06-27-121758_1600x900_scrot](https://cloud.githubusercontent.com/assets/10910745/16375549/981bb752-3c64-11e6-9853-bc7c915818c2.png)

#### Basic system details

MyPaint version: git revision 4275f1a
Operating system: Xubuntu 14.04
Desktop environment: Xfce

achadwick:
Reproduced with GNOME 3.20 and 4275f1a711af6cd6075c50ae9e37531b0fe17a74.

Yeah, gui.layers.RootStacklTreeView._button_press_cb() is horrible, messy code. The click detection aspects need a fairly thorough rewrite :scream:

PR #715, opened 2016-06-27
layers: Don't request to rename after double-clicking on expanders

danpla:
Fixes mypaint/mypaint#712.


achadwick:
I'm working on a more modular approach to #712 already :confused: 

It basically takes this approach for the double-click case.

danpla:
Ok, I'll close it.

PR #718, opened 2016-07-06
Use layer names instead of numbers when exporting multiple PNGs

timoschwarzer:
I think it is more useful to name the separate PNG files after its layers, instead of 001, 002 etc.

### DIFF
--- a/lib/document.py
+++ b/lib/document.py
@@ -1419,7 +1419,8 @@ class Document (object):
             prefix = l[0]
         doc_bbox = self.get_user_bbox()
         for i, l in enumerate(self.layer_stack.deepiter()):
-            filename = '%s.%03d%s' % (prefix, i+1, ext)
+            filename = '%s.%s%s' % (prefix, l.name, ext)
+
             l.save_as_png(filename, *doc_bbox, **kwargs)
 
     def load_png(self, filename, feedback_cb=None, **kwargs):


achadwick:
This is probably a good idea, but the implementation falls short of what's needed for general use.

Layer names cannot be used directly as filenames. Consider what happens if you have a `/` or a `\` or a `:` in a layer's name.

Layer names can be longer than file system limits. This code does not handle that or attempt to make anything shorter.

Names should also reflect layer order in the file somehow. The current `"%03d" % (i,)` thing is an acknowledgement of that.

Can you expand on your use case for this so we know a little more about why you want this?

timoschwarzer:
Yes, you are right. Totally missed these cases.

I'm currently using MyPaint to create game assets. And I use one Layer for one asset, so the resulting file contains a whole level. Because I use appropriate layer names for every asset, it's now easier to export them to multiple PNGs and know what's inside each file without looking at it. (if it's only Level1.019.png I don't know what's inside, but when it is Level1.boulder.png, it's more clear.)

0ion9:
* I'd suggest simply using a regexp substitution like `re.sub(r'[\\:/]+','_', layername)` to handle the first issue.

* Does this include path/hierarchy information? (eg. if you have a layer 'pineapple' in a group 'fruit', the output filename would reference both names, not just pineapple). I think that has the broadest application.

* I'm not sure about order. I think sometimes we will explicitly want it and other times explicitly not want it. If what you are doing is simply adding to a *set* of separate items, then your output is fundamentally unordered, and combining serial numbers with names actually will create file duplication whenever layers are moved around or added. Exporting parts of a composition is the opposite, of course.

* I'm tempted to say we should just make the user fix length problems (ie. check what length the output would be and refuse to export if the resulting filenames would be too long). My reason is that I believe making an effective abbreviator would be fairly complex, particularly given the suggestion to include hierarchy information; at minimum, adding an additional serial-number to layer names that would otherwise collide when abbreviated.

achadwick:
(Layers also have Python object id()s which don't change within a session, and UUIDs which I'd like to never change over the lifetime of a document. But we oughtn't to use those for naming: they're not especially human-relevant.)

There's also the **layer path**, often represented as something like "0:2:0:1" or the equivalent Python tuple of ints. It's a structural notation only, and will change as groups get moved around. It also isn't guaranteed to sort nicely, although it *might* in some file browsers. It's pretty terse though, and not many people will care about ID persistence when exporting - after all, the same is true of the serial ints used currently.

We can't hope to represent all parent layers in a filename. Too long.

Plenty of [slug](https://en.wikipedia.org/wiki/Semantic_URL#Slug)~ification techniques exist, for example: “Use layer names instead of numbers when exporting multiple PNGs” → `use-layer-names-instead-numbers-exporting-multiple-pngs`. They may be quite language-specific.

If strings might be truncated, we want to pack the most meaningful information up front.

Another use case: animators might want **highly predictable** names when exporting, if they're using tools which stitch together frames into movies.

0ion9:
> We can't hope to represent all parent layers in a filename. Too long.

If you mean that it's not possible, I disagree. 
* Taking the most restrictive OS (Windows, where the legacy API says paths must be <=260 bytes in total)
* Allocating 133 bytes to the directory part of the path, and 127 to the filename part of the path; the latter being broken into 30 characters for base filename, 97 for layer names.
* And using some multibyte unicode characters in layer names, we can fit an average of 5.8 levels of hierarchy in. (uniform random sampling [5..24] bytes per name). This includes accounting for separator characters, and does not truncate any names.

Code:

```
import random
results = []
ntrials = 100000
maxb = 127 - 30
for trial in range(ntrials):
    tmp = []
    while (sum(tmp) + len(tmp)) < maxb:
        tmp.append(random.randint(5,24))
    while (sum(tmp) + len(tmp)) > maxb:
        tmp.pop(-1)
    results.append(len(tmp))
sum(results) / len(results)
```

Anyway, that's just getting some data. On a practical level, I think there are some obvious uses for including *some* group names, as in my example.

* The simplest approach would probably to allow 'include N levels of hierarchy' to be configured per-document (counting backwards from the layer currently being exported, so for 'pineapple' layer, N=0 would result in 'pineapple', N=1 would result in 'fruit-pineapple', N=2 would result in 'icons-fruit-pineapple'..). I think this suits typical use cases but I don't have enough data to be sure yet.

* (another use case that is fairly typical IMO is to have a group per output item (each output item being comprised of, say, a mask layer, a shading layer, and a sfx layer). In this case you would want to output the group's compound content with the group's name, and not individually export the component layers. That might be achievable with a sort-of-opposite setting to the above, where increasing N ignores N levels of the hierarchy starting from the *deepest*.)

* combining that with an animation workflow, you might have a set of frames for one animation in one group, and another set for a different animation in a different group. So you might want jump-{000..010}, run-{000..005} ideally. The serial number might or might not be acceptable for this (same problem I described before, where you might end up with 'more' frames than expected due to reordering, adding, or deleting layers)

* so summing that up:
  * exporting groups rather than layers is useful
  * including some level of hierarchy information is useful, particularly if we can omit the higher-level parts of the the hierarchy
  * serial numbers may work for frame numbering, but have some gotchas (maybe it would be better to let the user rename a selection of layers using autonumbering?).

* IMO in all cases we should show a sample of the output filenames based on the currently selected layer.

odysseywestra:
@timoschwarzer could you update your pull request with @0ion9 and @achadwick's ideas and suggestions?

timoschwarzer:
I'm sorry but currently I don't have time to reinvestigate this.

odysseywestra:
Okay then. We'll probably label this one as stale and close it in a couple of weeks if no progress made. Or you could just close this PR, and then submit an issue ticket requesting this feature so we can look into it later.

Issue #529, opened 2015-12-11
Horizontal/snowflake symmetry


android272:
The current symmetry tool is very nice especially now that you can move it and delete the line.  But you can not make your image symmetrical horizontal or choose the angle of the line or add more than one line.  It would be really nice if you we had more lines of symmetry.

Horizontal symmetry would work exactly like vertical symmetry does now.  You select it and it mirrors what ever you draw on one side of the line to the other. You can also move the line and delete it as you can with vertical symmetry. the icon could be something like a mountain reflecting in a lake.

![Image of mountains](http://thumb7.shutterstock.com/photos/display_pic_with_logo/411121/411121,1256579078,1.jpg)

Then there is snowflake symmetry where you have many lines of symmetry.  I don't know how you would handle adding and subtracting more lines of symetry. The icon would be a snowflake.

![Image of snowflake](https://cdn0.iconfinder.com/data/icons/good-weather-1/96/weather_icons-68-512.png)

achadwick:
I're reported the brush engine aspects of this as mypaint/libmypaint#27. Although we may end up doing it in MyPaint itself if axes get very complex: perhaps one active brush state object per reflection, and one original. We'll have to see what's possible.

Even if libmypaint is updated to allow fancier symmetrical painting, there will still be user-interface aspects to implementing it. Those can go in this issue.

odysseywestra:
Closing since Pull Request #720 and mypaint/libmypaint#57 has addressed and added this Feature.

Shnatsel:
Is rotating the symmetry axis implemented in this PR? 

I could really use symmetry mode in the picture I'm currently working on, but the character's head is tilted so I need to rotate the symmetry axis.

odysseywestra:
As it now stands, that hasn't been added. @ShadowKyogre could you look into maybe adding this?

ShadowKyogre:
@Shnatsel @odysseywestra - Hey there! Sorry for the long wait for the reply. Was trying to get the hang of my new job before picking up coding again.

Anyway, I've worked out the math for supporting it in my graphics-playground repo.

Code for the raw math is here: https://github.com/ShadowKyogre/graphics-playground/blob/87ac39ba8b58524ce27213168d1bed653b0c9aaa/examples/slanted-reflect/cairo.py

Does this image reflect what you need? The arrow looking things represent hypothetical brush dabs to make sure I'm adjusting the angle correctly with the axis tilt.

[...]

PR #720, opened 2016-07-18
[RFC] Alternative symmetry modes

ShadowKyogre:
Implements #529 .

The reason why the [RFC] is in the pull request title is that I was working on getting this into a presentable, working state before asking for comments on the implementation.

The following commits implement horizontal, horizontal + vertical, rotational, and snowflake symmetry.

That being said, I'm fairly happy with how I implemented the UI. It's mostly straightforward.

I'm just confused at how I should handle the spinbox for the rotational symmetry lines. Without me adjusting the code for the alternative symmetry modes, the dialog box for mypaint's symmetry X axis isn't insensitive when symmetry is disabled. Is this intended? If not, I'll adjust the symmetry UI code to respond properly when symmetry is disabled.

Further commits here are likely going to be stylistic touches to adhere to the Python Style guide.

If pulled, it should only be pulled at the same time with https://github.com/mypaint/libmypaint/pull/57 to ensure the code builds.

### DIFF
--- a/gui/document.py
+++ b/gui/document.py
@@ -1723,16 +1723,19 @@ class Document (CanvasController):  # TODO: rename to "DocumentController"
         want_active = bool(action.get_active())
         if want_active and not already_active:
             alloc = self.tdw.get_allocation()
-            axis_pos = self.model.layer_stack.symmetry_axis
-            if axis_pos is None:
+            axis_x_pos = self.model.layer_stack.symmetry_x
+            axis_y_pos = self.model.layer_stack.symmetry_y
+            if axis_x_pos is None or axis_y_pos is None:
                 center_disp = alloc.width / 2.0, alloc.height / 2.0
                 center_model = self.tdw.display_to_model(*center_disp)
-                axis_pos = center_model[0]
-                self.model.layer_stack.symmetry_axis = axis_pos
+                axis_x_pos = center_model[0]
+                axis_y_pos = center_model[1]
+                self.model.layer_stack.symmetry_x = axis_x_pos
+                self.model.layer_stack.symmetry_y = axis_y_pos
         if want_active != already_active:
             self.model.layer_stack.symmetry_active = want_active
 
-    def _symmetry_state_changed_cb(self, layerstack, active, x):
+    def _symmetry_state_changed_cb(self, layerstack, active, x, y, sym_type, rot_sym_lines):
         """Update the SymmetryActive toggle on model state changes"""
         symm_toggle = self.action_group.get_action("SymmetryActive")
         symm_toggle_active = bool(symm_toggle.get_active())
--- a/gui/symmetry.py
+++ b/gui/symmetry.py
@@ -420,7 +420,7 @@ class SymmetryEditOptionsWidget (Gtk.Alignment):
         )
         self._axis_rot_symmetry_lines = Gtk.Adjustment(
             rootstack.rot_symmetry_lines,
-            upper=9,
+            upper=50,
             lower=2,
             step_incr=1,
             page_incr=3,
--- a/gui/symmetry.py
+++ b/gui/symmetry.py
@@ -23,6 +23,8 @@ import gui.tileddrawwidget
 import lib.alg
 from lib.color import RGBColor
 import lib.helpers
+import lib.mypaintlib
+import lib.tiledsurface
 import gui.drawutils
 from lib.gettext import C_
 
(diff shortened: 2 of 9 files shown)


odysseywestra:
@ShadowKyogre could you update both PR so we can do more testing?

odysseywestra:
@ShadowKyogre  I played around with it for a bit, and it looks great to start off. I love the option to add more symmetry tools to MyPaint. I do have a few suggestions and a question though.

Suggestions:
* An option to move both axis as one.
* An option to rotate the axis as a whole.
* An option to move an axis to the center of the view.
* An option to lock an axis line.

Question:
* Why is the rotational lines limited to 9?

I'm sure @achadwick would like to check on the code, but we need the Pull Request updated before we can move forward.

ShadowKyogre:
Rebased against master. IRL was pretty busy @_@. So, on the suggestions:

> * An option to move both axis as one.    
> * An option to move an axis to the center of the view.
> * An option to lock an axis line.

These should be implemented pretty quickly.

> * An option to rotate the axis as a whole.

This might be a bit tricky to do, but I'll take a poke around in the code to see how I can get the axes to rotate.

> * Why is the rotational lines limited to 9?

Mostly because I'm not sure what would be considered a sane number of rotational lines to limit it to. What would your suggestion be?

odysseywestra:
> Mostly because I'm not sure what would be considered a sane number of rotational lines to limit it to. What would your suggestion be?

Krita has it limited to 50 if you want a starting point.

ShadowKyogre:
@odysseywestra : Also, what's the most ideal position for moving both axis option? Right now, the delete axis overlay button is placed where the user would want to drag to move both axis. Where should that button be moved?

ShadowKyogre:
So the axes lock, logic wise, is almost fully implemented. It's just trying to figure out where to copy the toggle for the axes locks.

Here's my current draft of the commit along with the relevant line notes:

- https://ptpb.pw/OADN/diff#L-19 - Where the axes lock should be updated for the SymmetryEditMode
- https://ptpb.pw/OADN/diff#L-117 - Where the axes lock should be toggled from the options panel

Current candidates for passing the flag that I've considered, but haven't done so are:

- `tdw` doesn't look like the right place because that's TiledDrawWidget. That's a view, not user data.
- `tdw.doc` looks like the closest place because this might be data that the user wants to keep persistent for different pieces of artwork? I'm not entirely sure if I should be cramming it here, though.
- `app.preferences` doesn't look like the right place because this isn't something that'd be saved across different works.

What would you recommend as the place that the symmetry options panel should write to for the symmetry edit mode to pick up later?

odysseywestra:
Okay I got it to build and now I'm getting this error every time I try to switch back to freehand mode from the Symmetry tool:

```python
Mypaint version: None
System information: Linux-4.8.0-45-generic-x86_64-with-Ubuntu-16.10-yakkety
Using: Python 2.7.12, GTK 3.20.9, GdkPixbuf 2.34.0, Cairo 1.14.6, GLib 2.50.0
  File "/home/albert/Projects/Github/MyPaint/mypaint/build/lib.linux-x86_64-2.7/gui/tileddrawwidget.py", line 1159, _draw_cb(self=<tileddrawwidget.CanvasRenderer object at 0x7ff4...ileddrawwidget+CanvasRenderer at 0x5622c2ed2230)>, widget=<tileddrawwidget.CanvasRenderer object at 0x7ff4...ileddrawwidget+CanvasRenderer at 0x5622c2ed2230)>, cr=<cairo.Context object>)
                cr.save()
                overlay.paint(cr)
                cr.restore()
  variables: {'cr': ('local', <cairo.Context object at 0x7ff4bd0bd6f0>), 'overlay.paint': ('local', <bound method SymmetryOverlay.paint of <gui.symmetry.SymmetryOverlay object at 0x7ff4b8df1650>>)}
  File "/home/albert/Projects/Github/MyPaint/mypaint/build/lib.linux-x86_64-2.7/gui/symmetry.py", line 863, paint(self=<gui.symmetry.SymmetryOverlay object>, cr=<cairo.Context object>)
                if ax_point[0] == ax_point2[0]:
                    if mode.zone == _EditZone.MOVE_X_AXIS:
                        line_color = gui.style.ACTIVE_ITEM_COLOR
AttributeError: 'FreehandMode' object has no attribute 'zone'
```

odysseywestra:
> What would you recommend as the place that the symmetry options panel should write to for the symmetry edit mode to pick up later?

This is a question for @achadwick .

My two cents is where ever it applies the setting to the user config file and can be reset when a user deletes the config.

odysseywestra:
> Also, what's the most ideal position for moving both axis option? Right now, the delete axis overlay button is placed where the user would want to drag to move both axis. Where should that button be moved?

I would place the delete button in one of the corners where the user won't accidentally click on it. Or you can just keep them both in the center and just have the delete button move to the left or right of the Move button.

odysseywestra:
@ShadowKyogre I'm am going to Merge your PR as is. We can add the extra stuff later such as locking the axis line and centering them since they are not exactly critical to the feature. I'll make sure to add you to the @mypaint/developers team so you can push further commits that are related to the symmetry tool.

PR #724, opened 2016-08-15
py3k: use builtin next() function call

daingun:


achadwick:
Hmm, perhaps not. Spinbox widgets aren't iterators, just objects with a next()/previous() method pair. And I prefer to avoid using such things as if they were iterators.

QuLogic:
All the `next` should have been handled in #689 (unless they're new).

PR #728, opened 2016-08-21
Brush inputs/settings calibrated for viewzoom and viewrotation [WIP]

briend:
This is related to 
https://github.com/mypaint/libmypaint/pull/70
This adds viewzoom and viewrotation as inputs to libmypaint so the brush engine can make corrections to inputs to account for zoom level and canvas rotation.  Additional changes to the gui might be needed.

I just noticed the early commits have mention of pen-rotation-- this pull does not include pen-rotation.  I'm still learning git I must have goofed something there.

### DIFF
--- a/gui/drawutils.py
+++ b/gui/drawutils.py
@@ -40,14 +40,14 @@ from lib.pixbufsurface import render_as_pixbuf
 ## Module constants
 
 _BRUSH_PREVIEW_POINTS = [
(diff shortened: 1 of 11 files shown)


briend:
I think the checks are failing because it needs the new libmypaint here
https://github.com/mypaint/libmypaint/pull/70
sort of cart before the horse issue. . .

briend:
deleted and recreated this branch with a clean commit history

PR #731, opened 2016-08-27
Brush inputs/settings calibrated for viewzoom and viewrotation [WIP]

briend:
This is the GUI counterpart to libmypaint changes to correct brushes for view rotation and zoom:
https://github.com/mypaint/libmypaint/pull/70

### DIFF
--- a/gui/freehand.py
+++ b/gui/freehand.py
@@ -140,6 +140,8 @@ class FreehandMode (gui.mode.BrushworkModeMixin,
             self.last_good_raw_pressure = 0.0
             self.last_good_raw_xtilt = 0.0
             self.last_good_raw_ytilt = 0.0
+            self.last_good_raw_viewzoom = 0.0
+            self.last_good_raw_viewrotation = 0.0
 
         def queue_motion(self, event_data):
             """Append one raw motion event to the motion queue
@@ -148,7 +150,7 @@ class FreehandMode (gui.mode.BrushworkModeMixin,
             :type event_data: tuple
 
             Events are tuples of the form ``(time, x, y, pressure,
-            xtilt, ytilt)``. Times are in milliseconds, and are
+            xtilt, ytilt, viewzoom, viewrotation)``. Times are in milliseconds, and are
             expressed as ints. ``x`` and ``y`` are ordinary Python
             floats, and refer to model coordinates. The pressure and
             tilt values have the meaning assigned to them by GDK; if
@@ -157,7 +159,7 @@ class FreehandMode (gui.mode.BrushworkModeMixin,
 
             Zero-dtime events are detected and cleaned up here.
             """
-            time, x, y, pressure, xtilt, ytilt = event_data
+            time, x, y, pressure, xtilt, ytilt, viewzoom, viewrotation = event_data
             if time < self._last_queued_event_time:
                 logger.warning('Time is running backwards! Corrected.')
                 time = self._last_queued_event_time
@@ -166,7 +168,7 @@ class FreehandMode (gui.mode.BrushworkModeMixin,
                 # On Windows, GTK timestamps have a resolution around
                 # 15ms, but tablet events arrive every 8ms.
                 # https://gna.org/bugs/index.php?16569
-                zdata = (x, y, pressure, xtilt, ytilt)
+                zdata = (x, y, pressure, xtilt, ytilt, viewzoom, viewrotation)
                 self._zero_dtime_motions.append(zdata)
             else:
                 # Queue any previous events that had identical
@@ -182,9 +184,9 @@ class FreehandMode (gui.mode.BrushworkModeMixin,
                         zt = self._last_queued_event_time
                         interval = dtime
                     step = interval / (len(self._zero_dtime_motions) + 1)
-                    for zx, zy, zp, zxt, zyt in self._zero_dtime_motions:
+                    for zx, zy, zp, zxt, zyt, zvz, zvr in self._zero_dtime_motions:
                         zt += step
-                        zevent_data = (zt, zx, zy, zp, zxt, zyt)
+                        zevent_data = (zt, zx, zy, zp, zxt, zyt, zvz, zvr)
                         self.motion_queue.append(zevent_data)
                     # Reset the backlog buffer
                     self._zero_dtime_motions = []
@@ -290,6 +292,8 @@ class FreehandMode (gui.mode.BrushworkModeMixin,
             drawstate.last_good_raw_pressure = 0.0
             drawstate.last_good_raw_xtilt = 0.0
             drawstate.last_good_raw_ytilt = 0.0
+            drawstate.last_good_raw_viewzoom = 0.0
+            drawstate.last_good_raw_viewrotation = 0.0
 
             # Hide the cursor if configured to
             self._hide_drawing_cursor(tdw)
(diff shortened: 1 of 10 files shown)


odysseywestra:
@briend could you please update your pull requests so I can review it?

achadwick:
@odysseywestra Thanks for chasing these PRs ☺

@briend Can you collapse the whitespace-only commits into zero or one as well, please? A forced push onto this branch is fine for us.

I tend to go a step further and prefix sequences of changes with a Flake8-fixing pass, but that's not a MyPaint QA requirement... yet :smile:

odysseywestra:
@briend seems like your pull request is failing to build on both Tea-CI and Travis-CI. Could you check logs and see if you can correct the problems they are popping up?

briend:
I'm pretty sure they are failing because this patch needs the correlating patch to libmypaint, since I'm sending 2 more pieces of data.  I'm not sure how this can be resolved, chicken before the egg type problem?

odysseywestra:
Probably the best way to fix that would to have the CI tools to build with the Contributors Fork of `libmypaint` first. More information about this is on #784 .

lib/layer/data.py
```diff
@@ -1448,6 +1448,81 @@ def _load_strokemap_from_ora(self, elem, x, y, orazip=None, oradir=None):
...
+        new one.
+        """
+        self._surface.begin_atomic()
+        split = brush.stroke_to(
+            self._surface.backend, x, y,
+            pressure, xtilt, ytilt, dtime, viewzoom, viewrotation
+        )
+        self._surface.end_atomic()
+        self.autosave_dirty = True
+        return split
```
achadwick: Looks like like thee additions crept into the patch by accident during a rebase. These methods got moved into a new base class in master.

gui/freehand.py
```diff
@@ -359,6 +363,8 @@ def motion_notify_cb(self, tdw, event, fakepressure=None):
         pressure = event.get_axis(Gdk.AxisUse.PRESSURE)
         xtilt = event.get_axis(Gdk.AxisUse.XTILT)
         ytilt = event.get_axis(Gdk.AxisUse.YTILT)
+        viewzoom = tdw.scale
+        viewrotation = tdw.rotation
```
achadwick: Re the angle correction stuff for tilt at line 431 below (sorry, github doesn't allow me to comment there)

If we are capturing view rotation now and passing it to the brush engine, shouldn't that block be moved to the brush engine's stroke_to? The goal here is that brush authors get an ascension relative to the reference angle on the canvas, not the viewport. That needs to still be the case, but it would be better to move it into the brush engine at the same time we change its interface, given that we're changing what xtilt and ytilt mean.

With that in mind, shouldn't viewrotation be flipped too, if the view is mirrored? Same reason as in the correction calculation below ☺

achadwick:
@briend
I'd like to chase this one through because I want to fiddle with the freehand mode some more in order to repurpose its input-side cleanup for other things (e.g. selection editing and "rubber stamp" offset cloning, both of which involve drawing to out-of-tree layers... you'll see some conflicty evidence of that when you rebase, just to warn you, but it's just a single line)

The refactor I want to do is likely to break this patch again, so lets get your stuff into MyPaint first! :grinning:

briend:
Sorry about the duplicate lines, I still get confused by rebasing :-).  Do you think it makes sense to bundle this all with the barrel rotation stuff, since it touches all the same bits of code and is very tedious to merge?  I could make a new branch with view-rotation, zoom, and barrel-rotation all combined and cleaned up if you think that could help:
https://github.com/briend/libmypaint/tree/pen-rotation
https://github.com/briend/mypaint/tree/pen-rotation

achadwick:
Let's try doing it one by one so that the commits can be triaged more easily.

Do you think that using a glorified namedtuple "EventData" object in the Python code rather than all those separate values would make it more readable? I could add something like that to master first, and it'd give you a place to hang new inputs...

briend:
So we can reference them by names?  That would be awesome.  What about on the libmypaint side is there any chance to make that name based instead of position?  Also I wonder if there is any way to make it more lenient so that it's not fatal if you are missing an input or have an extra input?  I don't want to hold you up, feel free to make any changes and I can rework these branches.

gui/freehand.py
briend: I see what you mean; it would be better to let the brush engine take care as much as it can with the data it is given, right?  One thought I had is that if we ever want a virtual pen rendered along with dab angle (like some other programs do) we'd have to keep (or bring over more) stuff into the gui, duplicating the calculations?

Also, I wonder if other programs interested in the brush engine already have pre-calculated data that would be need to be converted back to "raw"?  I could imagine that being really annoying.

Would it make sense to somehow allow raw data OR calibrated data?  So the brush engine would just use either one.  I don't know how you could do that, but it'd be nice if you could name the data that you fed to the engine "raw_xtilt" vs "calibrated_xtilt", etc.  Then the engine could just have an if statement to handle them differently?

briend:
I fixed up both patches for this and pen-rotation and made pull requests, but I do need to test it a lot more in case I missed another stroke_to somewhere :-)

gui/freehand.py
achadwick: Yes, you're right with the other programs point. It's subtle and evil to change the meanings of tilt on them. Let's keep the meaning as at present.

The right place to hang this, if any optimization is needed, is `Brush::stroke_to()` in `brush.hpp`.

achadwick:
The right place to unbundle them from a Python object might be at the C++ wrapper layer, Brush::stroke_to(), before it hits libmypaint. Should be reasonably easy: for named access on a `namedtuple` or anything we implement, there's `PyObject_GetAttr()` in the cPython API.

achadwick:
Thanks for these! I notice that view rotation isn't exposed in the brush editor. Should it be?

I'd really appreciate a write-up of how to get a brush to have the same on-screen radius at every zoom level.  Could it be done in the libmypaint description strings?

achadwick:
@briend 
Sorry to be a pain. I've integrated the changes in this PR after satisfying myself that they are broadly working, since there's no way we could meaningfully test against libmypaint before. Several integrationny-looking bugs have tumbled out during testing:

* #818 (probably the most important)
* #817
* #816
* mypaint/libmypaint#89

I hope it's OK to assign these to you.

briend:
No trouble at all, I'm the one that broke stuff haha.  My afternoon is free I'll be able to concentrate on this today :)

briend:
I really wish it was more intuitive.  Here's some screenshots that should pretty much explain what's going on.

Here we are starting off with a big radius brush at 100% zoom.  I've dragged the zoom -4.15 so that it matches perfectly with the maximum zoom level.  I guess so you can say this is a 1:1 ratio of shrinking to zooming.  If you dragged it to -8 then the brush would shrink even faster than you are zooming (which might be useful).  Likewise, less than -4.15 would shrink slower than zoom.  

screenshot from 2017-06-17 13-37-49

Here I've zoomed in to 150% and drawn another dab.  It looks the same!  You can see the bigger dab to the left
screenshot from 2017-06-17 13-38-35

and here's 200%
screenshot from 2017-06-17 13-38-35

and even 550%
screenshot from 2017-06-17 13-39-16

ALL the way-- 6400% zoom.  You better adjust hardness and pixel feather (w/ zoom or manually) if you expect this range of zoom to work.

screenshot from 2017-06-17 14-26-15


and here's zoomed out at 37%
screenshot from 2017-06-17 13-39-42

There are limitations, if you zoom WAY you might hit the brush size limit (unless you started with a very small brush).  This seems acceptable since gigantic brushes aren't very useful.
screenshot from 2017-06-17 13-41-32

Issue #352, opened 2015-06-27
Add keyboard-friendly "reveal or hide" actions for panels like the Layers tree


achadwick:
I frequently find myself pressing "L" to see the layers dialog, then "L" again to hide it. This is not actually what I want, since the "hide" part of the toggle removes it from its sidebar.

I'd like instead for the current hide/show commands with default bindings to *reveal* the panel in its current location in the first instance, meaning to just open out its sidebar or show and present (raise) its floating window if it's floated out. If the sidebar is open or the floating window is visible, the key should close the sidebar holding the panel, or hide its floating window.

(This would work a lot more like the C/V/B colour and brush popups, in terms of user workflow. There may be some timing fussiness to get right though, since dockable panels would not appear under the pointer.)

These could be added as extra actions only accessible in the Keys tab of the preferences dialog.

PR #775, opened 2017-02-03
New Keybaord Shortcuts and Defaults

odysseywestra:
Added, changed, and edited the default shortcuts and tooltips for MyPaint. Also removed the RevealLayersTool, and similar objects and replaced them with their Toggle versions. Lastly, I've added the shortcuts for the Inking Tool.

### DIFF
--- a/gui/resources.xml
+++ b/gui/resources.xml
@@ -167,6 +167,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Try to save and resume interrupted unsaved work.</property>
           <signal name="activate" handler="autorecover_cb"/>
         </object>
+        <accelerator key="F1"/>
       </child>
     </object>
   </child>
@@ -210,7 +211,7 @@ Vocabulary
           <!-- Code in document.py takes care of ltr/rtl details -->
           <signal name="activate" handler="undo_cb"/>
         </object>
-        <accelerator key="z"/>
+        <accelerator key="z" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="Redo">
@@ -218,7 +219,7 @@ Vocabulary
           <!-- Code in document.py takes care of ltr/rtl details -->
           <signal name="activate" handler="redo_cb"/>
         </object>
-        <accelerator key="y"/>
+        <accelerator key="z" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
       </child>
       <!--}}} -->
       <!--{{{ Brush modifications -->
@@ -263,6 +264,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Increase the lightness of the current painting color.</property>
           <signal name="activate" handler="brighter_cb"/>
         </object>
+        <accelerator key= "f" modifiers="GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="Darker">
@@ -270,6 +272,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Decrease the lightness of the current painting color.</property>
           <signal name="activate" handler="darker_cb"/>
         </object>
+        <accelerator key="d" modifiers="GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="IncreaseHue">
@@ -291,6 +294,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Make the painting color more saturated (more colorful, purer).</property>
           <signal name="activate" handler="purer_cb"/>
         </object>
+        <accelerator key="s" modifiers="GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="Grayer">
@@ -298,6 +302,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Make the painting color less saturated (grayer).</property>
           <signal name="activate" handler="grayer_cb"/>
         </object>
+        <accelerator key="a" modifiers="GDK_SHIFT_MASK"/>
       </child>
       <!-- Resetting -->
       <child>
@@ -307,13 +312,14 @@ Vocabulary
           <signal name="activate" handler="brush_reload_cb"/>
           <property name="icon-name">mypaint-clear-all-symbolic</property>
         </object>
+        <accelerator key="grave"/>
       </child>
       <!-- }}} -->
       <!--{{{ Picking strokes -->
       <child>
         <object class="GtkAction" id="PickContext">
           <property name="label" translatable="yes" context="Menu→Brush (labels), Accel Editor (labels)">Pick Stroke and Layer</property>
-          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Pick a brushstroke from the canvas: restores brush, color, and layer.</property>
+          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Pick a brushstroke from the canvas which restores brush, color, and layer.</property>
           <property name="icon-name">mypaint-target-symbolic</property>
           <signal name="activate" handler="pick_context_cb"/>
         </object>
@@ -328,6 +334,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Toggle whether restoring the brush from a shortcut key also restores the saved color.</property>
           <signal name="activate" handler="context_toggle_color_cb"/>
         </object>
+        <accelerator key="asciitilde"/>
       </child>
       <child>
         <object class="GtkAction" id="ContextStore">
@@ -343,7 +350,7 @@ Vocabulary
         <object class="GtkAction" id="ClearLayer">
           <property name="icon-name">mypaint-clear-all-symbolic</property>
           <property name="label" translatable="yes" context="Menu→Layer (labels), Accel Editor (labels)">Clear Layer</property>
-          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Clear the current layer’s content.</property>
+          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Clear the active layer’s content.</property>
           <signal name="activate" handler="clear_layer_cb"/>
         </object>
         <accelerator key="Delete"/>
@@ -354,6 +361,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Erase parts of the current layer which lie outside the frame.</property>
           <signal name="activate" handler="trim_layer_cb"/>
         </object>
+        <accelerator key="t" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="CopyLayer">
@@ -379,6 +387,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Begin editing the current layer in an external application.</property>
           <signal name="activate" handler="begin_external_layer_edit_cb"/>
         </object>
+        <accelerator key="l" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="CommitExternalLayerEdit">
@@ -390,7 +399,7 @@ Vocabulary
       <child>
         <object class="GtkAction" id="PickLayer">
           <property name="label" translatable="yes" context="Menu→Layer→Go to Layer (labels), Accel Editor (labels)">Go to Layer at Cursor</property>
-          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Select a layer by picking an object on it.</property>
+          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Select a layer by picking an object from the layer.</property>
           <signal name="activate" handler="pick_layer_cb"/>
         </object>
         <accelerator key="h"/>
@@ -437,6 +446,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Add a new layer group above the current layer.</property>
           <signal name="activate" handler="new_layer_cb"/>
         </object>
+        <accelerator key="g" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="NewPaintingLayerBelow">
@@ -459,6 +469,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Add a new layer group below the current layer.</property>
           <signal name="activate" handler="new_layer_cb"/>
         </object>
+        <accelerator key="g" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="MergeLayerDown">
@@ -467,7 +478,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Merge the current layer with the layer beneath it.</property>
           <signal name="activate" handler="merge_layer_down_cb"/>
         </object>
-        <accelerator key="Delete" modifiers="GDK_CONTROL_MASK"/>
+        <accelerator key="m" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="MergeVisibleLayers">
@@ -475,6 +486,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Consolidate all visible layers, and delete the originals.</property>
           <signal name="activate" handler="merge_visible_layers_cb"/>
         </object>
+        <accelerator key="Delete" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="NewLayerMergedFromVisible">
@@ -482,6 +494,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Like ‘Merge Visible Layers’, but doesn’t delete the originals.</property>
           <signal name="activate" handler="new_layer_merged_from_visible_cb"/>
         </object>
+        <accelerator key="m" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="RemoveLayer">
@@ -490,7 +503,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Remove the current layer from the layer stack.</property>
           <signal name="activate" handler="remove_layer_cb"/>
         </object>
-        <accelerator key="Delete" modifiers="GDK_SHIFT_MASK"/>
+        <accelerator key="Delete" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="NormalizeLayerMode">
@@ -498,6 +511,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Convert the current layer or group to a painting layer in ‘Normal’ mode, preserving its appearance.</property>
           <signal name="activate" handler="normalize_layer_mode_cb"/>
         </object>
+        <accelerator key="n" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="IncreaseLayerOpacity">
@@ -522,6 +536,7 @@ Vocabulary
           <property name="icon-name">mypaint-layer-raise-symbolic</property>
           <signal name="activate" handler="reorder_layer_cb"/>
         </object>
+        <accelerator key="Page_Up" modifiers="GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="LowerLayerInStack">
@@ -530,6 +545,7 @@ Vocabulary
           <property name="icon-name">mypaint-layer-lower-symbolic</property>
           <signal name="activate" handler="reorder_layer_cb"/>
         </object>
+        <accelerator key="Page_Down" modifiers="GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="DuplicateLayer">
@@ -538,6 +554,7 @@ Vocabulary
           <property name="icon-name">mypaint-duplicate-symbolic</property>
           <signal name="activate" handler="duplicate_layer_cb"/>
         </object>
+        <accelerator key="d" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="RenameLayer">
@@ -545,6 +562,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Enter a new name for the current layer.</property>
           <signal name="activate" handler="rename_layer_cb"/>
         </object>
+        <accelerator key="r" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkToggleAction" id="LayerLockedToggle">
@@ -552,6 +570,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Toggle the current layer’s locked state. Locked layers cannot be modified.</property>
           <signal name="activate" handler="layer_lock_toggle_cb"/>
         </object>
+        <accelerator key="End" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkToggleAction" id="LayerVisibleToggle">
@@ -559,6 +578,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Toggle the current layer’s visibility state, making it hidden.</property>
           <signal name="activate" handler="layer_visible_toggle_cb"/>
         </object>
+        <accelerator key="Home" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkToggleAction" id="ShowBackgroundToggle">
@@ -566,6 +586,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Toggle the background layer’s visibility.</property>
           <signal name="activate" handler="show_background_toggle_cb"/>
         </object>
+        <accelerator key="b" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
       </child>
       <!-- }}} -->
       <!-- {{{ View manipulation -->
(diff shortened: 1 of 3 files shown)


gui/resources.xml
```diff
@@ -490,14 +503,15 @@ Vocabulary
...
-        <accelerator key="Delete" modifiers="GDK_SHIFT_MASK"/>
+        <accelerator key="Delete" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="NormalizeLayerMode">
           <property name="label" translatable="yes" context="Menu→Layer (labels), Accel Editor (labels)">Convert to Normal Painting Layer</property>
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Convert the current layer or group to a painting layer in ‘Normal’ mode, preserving its appearance.</property>
           <signal name="activate" handler="normalize_layer_mode_cb"/>
         </object>
+        <accelerator key="n" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
```
achadwick: I think this needs a use case for a keyboard shortcut, even a heavily modified one. I'd be happy leaving it as an advanced feature which can only be engaged by visiting the Layers panel or the menus.

(Normalize is weird and finicky, because it sometimes picks up background elements if those are made to look different when you normalize a layer. Users can be surprised by this because it may not be immediately apparent. Let's make advanced things that have to be thought about specially _look_ rare and advanced by not assigning them keys (although all that meta-alt-ctrl-shift-splarg-doubledeclutch-N stuff serves much the same purpose tbh!))

gui/resources.xml
```diff
@@ -530,6 +545,7 @@ Vocabulary
           <property name="icon-name">mypaint-layer-lower-symbolic</property>
           <signal name="activate" handler="reorder_layer_cb"/>
         </object>
+        <accelerator key="Page_Down" modifiers="GDK_SHIFT_MASK"/>
```
achadwick: :+1: We've needed a shortcut for these for ages.

gui/resources.xml
```diff
@@ -538,34 +554,39 @@ Vocabulary
...
         </object>
+        <accelerator key="d" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="RenameLayer">
           <property name="label" translatable="yes" context="Menu→Layer (labels), Accel Editor (labels)">Rename Layer…</property>
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Enter a new name for the current layer.</property>
           <signal name="activate" handler="rename_layer_cb"/>
         </object>
+        <accelerator key="r" modifiers="GDK_CONTROL_MASK"/>
```
achadwick: Just a thought - GIMP uses F2 for this, and I'm forever finding myself saving scraps in MyPaint as a result, and messing up my session. Maybe it's time to revisit scraps navigation too?

gui/resources.xml
```diff
@@ -597,21 +618,21 @@ Vocabulary
         <object class="GtkAction" id="ResetZoom">
           <property name="icon-name">mypaint-view-100-symbolic</property>
           <property name="label" translatable="yes" context="Menu→View→Reset (labels), Accel Editor (labels)">Reset Zoom</property>
-          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Restore the view’s zoom to its default.</property>
+          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Restore the view’s zoom to 100%.</property>
```
achadwick: :-1: The default can be something other than 100%.

gui/resources.xml
```diff
@@ -597,21 +618,21 @@ Vocabulary
...
           <signal name="activate" handler="reset_view_cb"/>
         </object>
       </child>
       <child>
         <object class="GtkAction" id="ResetRotation">
           <property name="label" translatable="yes" context="Menu→View→Reset (labels), Accel Editor (labels)">Reset Rotation</property>
-          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Restore the view’s rotation to 0°</property>
+          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Reset the view’s rotation to 0°</property>
```
achadwick: :+1: I think it was intended as a gloss for then slightly-technical word "Reset", before the accelerator editor came along. Now it's unhelpful wording because it's not clear that it means what the menu calls "Reset".

gui/resources.xml
```diff
@@ -597,21 +618,21 @@ Vocabulary
...
           <signal name="activate" handler="reset_view_cb"/>
         </object>
       </child>
       <child>
         <object class="GtkAction" id="ResetMirror">
           <property name="label" translatable="yes" context="Menu→View→Reset (labels), Accel Editor (labels)">Reset Mirroring</property>
-          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Turn off mirroring of the view.</property>
+          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Reset the view's mirroring to its original orientation.</property>
```
achadwick: :-1: Drop the "its original orientation" bit.

:+1: On the "reset the view" wording.

gui/resources.xml
```diff
@@ -763,27 +784,35 @@ Vocabulary
       <!-- {{{ inktool points related actions -->
       <child>
         <object class="GtkAction" id="InsertCurrentNode">
-          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Insert a node back of the currently selected node.</property>
+          <property name="label" translatable="yes" context="Accel Editor (labels)">Insert Inking Node</property>
+          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Insert a inking node on the back of the currently selected node.</property>
           <signal name="activate" handler="insert_current_node_cb"/>
         </object>
+        <accelerator key="Insert"/>
```
achadwick: :-1: This is only relevant in Inking mode, and it doesn't have a menu entry, so it doesn't need a shortcut. let's leave it open till we add a generic "insert a thing on the current layer" action that activates this one if you're currently inking.

gui/resources.xml
```diff
@@ -763,27 +784,35 @@ Vocabulary
       <!-- {{{ inktool points related actions -->
       <child>
         <object class="GtkAction" id="InsertCurrentNode">
-          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Insert a node back of the currently selected node.</property>
+          <property name="label" translatable="yes" context="Accel Editor (labels)">Insert Inking Node</property>
+          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Insert a inking node on the back of the currently selected node.</property>
```
achadwick: The button calls it "Insert Point". "Node" is a bit technical.

"[In/on the] back of" isn't comprehensible to British English speakers, and I think it's factually incorrect too if it means "behind"! The node is inserted after the selected node, in timewise/strokewise ordering (= towards the tick mark)

gui/resources.xml
```diff
@@ -763,27 +784,35 @@ Vocabulary
...
         </object>
+        <accelerator key="Insert" modifiers="GDK_SHIFT_MASK | GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="CullNodes">
+          <property name="label" translatable="yes" context="Accel Editor (labels)">Cull Inking Node</property>
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Cull inking tool nodes alternately.</property>
           <signal name="activate" handler="cull_nodes_cb"/>
         </object>
+        <accelerator key="Insert" modifiers="GDK_CONTROL_MASK | GDK_MOD1_MASK"/>
```
achadwick: See the comment above. Currently they're only accessible via the panel, which is why they don't have labels.

It's fine to add labels though - they just won't be seen!

"Simplify Ink Points"?

gui/resources.xml
```diff
@@ -834,7 +863,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Apply color to the current layer without affecting its brightness or opacity.</property>
           <signal name="activate" handler="blend_mode_colorize_cb"/>
         </object>
-        <accelerator key="k" modifiers="GDK_SHIFT_MASK"/>
+        <accelerator key="e" modifiers="GDK_SHIFT_MASK"/>
```
achadwick: :+1: Probably a good idea to pack these oddities on single related key.

gui/resources.xml
```diff
@@ -995,27 +1028,31 @@ Vocabulary
...
         </object>
+        <accelerator key="r" modifiers="GDK_CONTROL_MASK | GDK_MOD1_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="ScratchSaveAsDefault">
           <property name="label" translatable="yes" context="Menu→Scratchpad (labels), Accel Editor (labels)">Save Scratchpad as Default</property>
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Save the scratchpad as the default for ‘New Scratchpad’.</property>
           <signal name="activate" handler="save_scratchpad_as_default_cb"/>
         </object>
+        <accelerator key="d" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK | GDK_MOD1_MASK"/>
```
achadwick: :-1: I don't think we should have any shortcuts for these secondary functions of the scratchpad.

gui/resources.xml
```diff
@@ -995,27 +1028,31 @@ Vocabulary
...
       </child>
       <child>
         <object class="GtkAction" id="ScratchClearDefault">
           <property name="label" translatable="yes" context="Menu→Scratchpad (labels), Accel Editor (labels)">Clear Default Scratchpad</property>
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Clear the default scratchpad to a blank canvas.</property>
           <signal name="activate" handler="clear_default_scratchpad_cb"/>
         </object>
+        <accelerator key="c" modifiers="GDK_CONTROL_MASK | GDK_MOD1_MASK"/>
```
achadwick: :-1: I don't think we should have any shortcuts for these secondary functions of the scratchpad.

gui/resources.xml
```diff
@@ -995,27 +1028,31 @@ Vocabulary
...
       </child>
       <child>
         <object class="GtkAction" id="ScratchCopyBackground">
           <property name="label" translatable="yes" context="Menu→Scratchpad (labels), Accel Editor (labels)">Copy Main Background to Scratchpad</property>
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Copy the background image from the working document into the Scratchpad.</property>
           <signal name="activate" handler="scratchpad_copy_background_cb"/>
         </object>
+        <accelerator key="b" modifiers="GDK_SHIFT_MASK"/>
```
achadwick: :-1: I don't think we should have any shortcuts for these secondary functions of the scratchpad.

gui/resources.xml
```diff
@@ -1104,6 +1141,7 @@ Vocabulary
           <property name="icon-name">mypaint-info-symbolic</property>
           <signal name="activate" handler="show_online_help_cb"/>
         </object>
+        <accelerator key="F1" modifiers="GDK_SHIFT_MASK"/>
```
achadwick: :+1: Though arguably we should review the scrap browser functions and use F1 for plain help.

gui/resources.xml
```diff
@@ -1062,7 +1099,7 @@ Vocabulary
           <property name="icon-name">mypaint-color-symbolic</property>
           <signal name="activate" handler="quick_chooser_popup_cb"/>
         </object>
-        <accelerator key="c"/>
+        <accelerator key="v" modifiers="GDK_SHIFT_MASK"/>
```
achadwick: Okay, in combination with the other changes for XCV.

gui/resources.xml
```diff
@@ -1173,7 +1211,7 @@ Vocabulary
       </child>
       <child>
         <object class="GtkAction" id="ShowPopupMenu">
-          <property name="label" translatable="yes" context="Menu→View (labels), Accel Editor (labels)">Popup Menu</property>
+          <property name="label" translatable="yes" context="Menu→View (labels), Accel Editor (labels)">Popup Main Menu</property>
```
achadwick: :+1:, we now have secondary menus too, but this key always triggers the main one. So it needs to say so :)

gui/resources.xml
```diff
@@ -1234,19 +1272,11 @@ Vocabulary
       <!-- {{{ Layers panel -->
       <child>
         <object class="GtkToggleAction" id="ToggleLayersTool">
-          <property name="label" translatable="yes" context="Menu→Window (labels), Toolbar (labels), Accel Editor (labels)">Layers Panel</property>
-          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Toggle the Layers dockpanel (arrange layers, or assign modes).</property>
+          <property name="label" translatable="yes" context="Menu→Layers (lables), Menu→Window (labels), Toolbar (labels), Accel Editor (labels)">Layers Panel</property>
```
achadwick: Typo ☺

gui/resources.xml
```diff
@@ -1234,19 +1272,11 @@ Vocabulary
...
           <signal name="activate" handler="toggle_dockpanel_cb"/>
         </object>
-      </child>
-      <child>
-        <object class="GtkAction" id="RevealLayersTool">
-          <property name="label" translatable="yes" context="Menu→Layers (labels), Accel Editor (labels)">Show Layers Panel</property>
-          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Reveal the Layers dockpanel (arrange layers, or assign modes).</property>
-          <property name="icon-name">mypaint-layers-symbolic</property>
-          <signal name="activate" handler="reveal_dockpanel_cb"/>
-        </object>
```
achadwick: :-1: :-1: :-1:
I really want to keep the reveal commands, because replacing them with toggle commands changes the sidebar order. See #352 for the background to this.

If they are not defined, things will break or code will become dead and thus a burden to maintain.

It may be fine to move this one to

gui/resources.xml
```diff
@@ -1670,6 +1673,7 @@ Vocabulary
           <property name="icon-name">mypaint-layer-move-symbolic</property>
           <signal name="activate" handler="mode_flip_action_activated_cb"/>
         </object>
+        <accelerator key="m"/>
```
odysseywestra: This on I was actually mimicking Gimps default key for move tool.

gui/resources.xml
```diff
@@ -1650,6 +1652,7 @@ Vocabulary
           <property name="icon-name">mypaint-tool-inking-symbolic</property>
           <signal name="activate" handler="mode_flip_action_activated_cb"/>
         </object>
+        <accelerator key="y"/>
```
odysseywestra: Do you have an alternative key for Inking tool then? I was thinking `<accelerator key="i"/> but that would mean I would need to move the MirrorView keys.

gui/resources.xml
```diff
@@ -1289,7 +1319,7 @@ Vocabulary
           <property name="icon-name">mypaint-tool-hsvsquare</property>
           <property name="label" translatable="yes" context="Menu→Color (labels), Accel Editor (labels)">HSV Square</property>
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Dockable color selector: HSV square which can be rotated to show different hues.</property>
-          <!-- FIXME: this description needs to be improved -->
+          <!-- FIXME: this description needs to be improved Note-ALbert: Looks good to me.-->
```
odysseywestra: Then how about " Dockable color selector: HSV square that has a ring around which can be rotated to show different hues."?

gui/resources.xml
odysseywestra: Okay then how do we hide it from being shown in the Accelerator Config. Map?

gui/resources.xml
odysseywestra: Okay would "Reset view's zoom level to its default." work?

gui/resources.xml
odysseywestra: Krita uses "F2" as well. What would you suggest to move the scraps navigation to then?

gui/resources.xml
odysseywestra: Yeah, thats why assigned the modifiers to it, however you are right. If the Normalize layer is and advance option, then it's probably better to have the user manually select it. Reverting the change.

gui/resources.xml
```diff
@@ -210,15 +211,15 @@ Vocabulary
           <!-- Code in document.py takes care of ltr/rtl details -->
           <signal name="activate" handler="undo_cb"/>
         </object>
-        <accelerator key="z"/>
+        <accelerator key="z" modifiers="GDK_CONTROL_MASK"/>
```
odysseywestra: Yeah, that would probably be a good idea to to have the Z and Y key to be in `gui/document.py` for backwards compatibility sake. Then we could move the Ctrl+Z and Ctrl+Shift Z from `gui/document.py` to here in the `resources.xml` instead.

gui/resources.xml
```diff
@@ -291,13 +294,15 @@ Vocabulary
...
         </object>
+        <accelerator key="s" modifiers="GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="Grayer">
           <property name="label" translatable="yes" context="Menu→Color→Adjust Color (labels), Accel Editor (labels)">Decrease Saturation</property>
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Make the painting color less saturated (grayer).</property>
           <signal name="activate" handler="grayer_cb"/>
         </object>
+        <accelerator key="a" modifiers="GDK_SHIFT_MASK"/>
```
odysseywestra: Oh. I like the idea of using those two for rotating the Hue. I will mark that down in my notes.

gui/resources.xml
```diff
@@ -459,6 +469,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Add a new layer group below the current layer.</property>
           <signal name="activate" handler="new_layer_cb"/>
         </object>
+        <accelerator key="g" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
```
odysseywestra: Yeah, your right. When I create a group layer I always tend to make it above the active layer than below it. I'll revert it.

gui/resources.xml
```diff
@@ -467,21 +478,23 @@ Vocabulary
...
         </object>
+        <accelerator key="Delete" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="NewLayerMergedFromVisible">
           <property name="label" translatable="yes" context="Menu→Layer (labels), Accel Editor (labels)">New Layer from Visible</property>
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Like ‘Merge Visible Layers’, but doesn’t delete the originals.</property>
           <signal name="activate" handler="new_layer_merged_from_visible_cb"/>
         </object>
+        <accelerator key="m" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
```
odysseywestra: No not really. The reason I would have a shortcut for this is to allow me to test what the layer would look like before I commit to merging the layers.

Though now that you say that, I think having a `MergeGroup` layer option would be nice. Plus having `NewGroupFromVisible` tool. I'll have to look into that in the future.

So I'm guessing since it is an advance feature(like the Normalize Layer) it should be up to the user to manually click it in the menu?

gui/resources.xml
odysseywestra: Using F1 for help is part of Gnome's HIG so we should look into moving the Scrap Browser functions.

gui/resources.xml
odysseywestra: Your right, it got confusing for me too. Reverting.

odysseywestra:
@achadwick thank for your review comments, and also sharing the [Gnome HIG](https://developer.gnome.org/hig-book/3.2/input-keyboard.html.en) page. I will definitely keep those in mind when going through the shortcut list again. I will also leave comments on why I made the changes. The comments will be removed them before the Pull Request is merge.

odysseywestra:
Also could you look into why tea-ci keeps failing on the 32-bit end?

gui/resources.xml
0ion9: I'm a bit confused about why the word 'its' -- which makes the whole phrase awkward -- is in there. We can't reset the zoom of the scratchpad using this action, and it also suggests that individual views may have different defaults (which would be a Bad Idea. Thankfully we aren't currently doing it.)
.. so can't we just say 'Reset view zoom level to default', no 'its'?

gui/resources.xml
0ion9: Why would you hide it? It's much more useful than the *toggle* for the layers panel is.

gui/resources.xml
0ion9: Is the fact that the ring is rotatable important, or could we just say 'HSV square with Hue ring'?

odysseywestra: Yeah, the ring part is important since it the part that can change hue of the color.

gui/resources.xml
```diff
@@ -379,6 +387,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Begin editing the current layer in an external application.</property>
           <signal name="activate" handler="begin_external_layer_edit_cb"/>
         </object>
+        <accelerator key="l" modifiers="GDK_CONTROL_MASK"/>
```
odysseywestra: If we were to move the scrap navigation shortcuts, couldn't we use one of those function keys? The reason I'm saying is that none of the other graphics like Gimp or Krita has this feature. So this is a key difference between the rest of the pack which should be highlighted and given a default shortcut.

gui/resources.xml
```diff
@@ -328,6 +334,7 @@ Vocabulary
           <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Toggle whether restoring the brush from a shortcut key also restores the saved color.</property>
           <signal name="activate" handler="context_toggle_color_cb"/>
         </object>
+        <accelerator key="asciitilde"/>
```
odysseywestra: I see.. cause my keyboard (US) the key is on the same row as the top row number keys. ^^'

gui/resources.xml
odysseywestra: That'll work.

gui/resources.xml
```diff
@@ -390,7 +399,7 @@ Vocabulary
       <child>
         <object class="GtkAction" id="PickLayer">
           <property name="label" translatable="yes" context="Menu→Layer→Go to Layer (labels), Accel Editor (labels)">Go to Layer at Cursor</property>
-          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Select a layer by picking an object on it.</property>
+          <property name="tooltip" translatable="yes" context="Accel Editor (descriptions)">Select a layer by picking an object from the layer.</property>
```
odysseywestra: Would "Select a layer by picking a brush stroke from the layer." work?

gui/resources.xml
odysseywestra: Technically speaking, how is it any better than the toggle? They essentially do the same thing expect that the toggle version allows you use the same key to reveal and hide the layer panel? To me having the reveal panel option available on the UI end is redundant.

gui/resources.xml
odysseywestra: What about "A dockable HSV color selector which focuses the square's value and saturation range based on the position of the ring around the square."? Even though it is a mouthful, that dose explain it. 

We could probably go with @0ion9 suggestion, but we probably want to word the other HSV selectors with the same structure whichever direction we go.

gui/resources.xml
0ion9: They don't do the same thing. The toggle version shuffles the layout.
You may regard it as a trivial difference, but from my point of view, that single difference makes 'Toggle layer panel' an action I never want to invoke. 

Being a little more objective and focusing on use cases: I see no reason why either
a) a user who has currently fullscreened MyPaint would want to invoke Toggle rather than Reveal
b) a user who is in non-fullscreen mode would want to invoke Reveal rather than Toggle.

It would be nice to have an action that *just worked* for all cases. But I don't think there is any realistic argument to be made that we currently have such an action.

odysseywestra: Ahhh that makes much more sense. So I will need to clarify the tooltips for the reveal versions that they are mainly used in fullscreen mode than and don't reset window.

This is a really confusing way to handle the panels cause what works with one will not work with the other and vice a versa.  Probably need to add a FIXME tag so it can be addressed in the future. I will revert the reveal version to the default keys, but I will also need to assigned the same keys for the toggle, but instead have a modifier attached to them instead.

Thanks you for clarifying that @0ion9. ^^

gui/resources.xml
achadwick: "Dockable color selector: HSV square with Hue ring" seems like the best compromise.

(I'm mostly whining about coparing it to the old HSV triangle, where the triangular bit really did get rotated)

gui/resources.xml
achadwick: /nods
Yeah, fullscreen mode is quite a different workflow sometimes, but it really is two states we're talking about here. "Present in the UI" and "Currently visible in the UI".

Tricky to simplify that sort of fine-grained distinction for presenting to a user, however. I think I decided that the Toggle variant was better to shove into a standard-looking Window menu while having the Reveal variant present near the top of the related menu to let users quickly find their Layers panel (say) if it got buried under a ton of other tabs.

gui/resources.xml
achadwick: Ctrl- variants sound right for the scrapbook functions.

Sadly, there's not really any standard for media-function keys on laptops we could leverage (main laptop :rewind:=Fn+F8 :play_or_pause_button:=Fn+F9, :fast_forward:=Fn+F10, windows clunker uses F9/F10/F11 for the same)

It frees up F2 for "rename current layer", matching GIMP :smiley:

gui/resources.xml
achadwick: Works for me too :)

gui/resources.xml
achadwick: I think keeping them on their current keys, but Ctrl-modified, will work best.

gui/resources.xml
achadwick: "Select a layer by picking something [drawn] on it" is probably just vague enough to fit all cases☺

gui/resources.xml
odysseywestra: but adding a control modifier to it wouldn't hurt. It would free it up for any future drawing tools modes.

gui/resources.xml
odysseywestra: If I add a modifier to the move layer tool. We could assign the M key to the inking tool.

gui/resources.xml
odysseywestra: The only problem with that is that is a global shortcut depending on the DE so it will interfere with a music player sadly.

I did find out that F6, F7, F8 are free though.

odysseywestra:
Okay new commit submitted. Here is a link to the app image that was made in Travis for testing.
https://transfer.sh/AieTe/mypaint-glibc2.15-x86-64.appimage

odysseywestra:
@mypaint/developers  I'm ready for another round of review.

gui/resources.xml
```diff
@@ -210,15 +211,15 @@ Vocabulary
...
+        <accelerator key="z" modifiers="GDK_CONTROL_MASK"/>
       </child>
       <child>
         <object class="GtkAction" id="Redo">
           <property name="icon_name">mypaint-redo-ltr-symbolic</property>
           <!-- Code in document.py takes care of ltr/rtl details -->
           <signal name="activate" handler="redo_cb"/>
         </object>
-        <accelerator key="y"/>
+        <accelerator key="z" modifiers="GDK_CONTROL_MASK | GDK_SHIFT_MASK"/>
```
martinxyz: Hm, Ctrl-Y for redo is quite common I think? What I wanted at the time was to reach all common features with a single finger, that's why I removed the Ctrl. Two-key combos are a bit more difficult to do single-handed.

achadwick: On reflection, Ctrl+Y is common too - for example, in this very text input box.

Maybe GNOME are being a bit too perfectionist again. On the other claw, Ctrl+Y is overloaded for paste when in Emacs input mode. Whatever that is.

Maybe Ctrl+Shift+Z could be the fallback, Ctrl+Y could be the advertised Undo binding?

odysseywestra: I rather stick with what the HIG suggests and leave the others(Z,Y, Ctrl+Y) as fallbacks. Regardless where the defaults are, all three sets are going to work in MyPaint out of the box. This will at least give the user the option to reassign the Z, Y, and Ctrl+Y if they wish without messing with the initial sets. I can at least add another phrase to the tooltips that Z, Y, and Ctrl+Y work as well so the user is aware of it.

odysseywestra:
@achadwick just checking in to see if you or anyone from @mypaint/developers have anymore comments on this one. If not I think this one is ready to be merged.

odysseywestra:
Just FYI I'm going to merge this PR Monday, unless one someone in the @mypaint/developers suggests a change.

PR #789, opened 2017-03-02
Mode stack: Add layer for brush tools

ydahhrk:
MyPaint now remembers the last used brush-based tool, so when the user chooses a new brush (via a shortcut or whatever), the active tool becomes something relevant.

### DIFF
--- a/gui/brushmanager.py
+++ b/gui/brushmanager.py
@@ -34,6 +34,8 @@ from lib.observable import event
 import lib.pixbuf
 import drawutils
 
+import gui.mode
+
 
 ## Public module constants
 
@@ -181,6 +183,10 @@ class BrushManager (object):
 
         # Update the history at the end of each definite input stroke.
         self.app.doc.input_stroke_ended += self._input_stroke_ended_cb
+
+        # Make sure the user always gets a brush tool when they pick a brush
+        # preset.
+        self.brush_selected += self._brush_selected_cb
 
     def _load_brush(self, brush_cache, name, **kwargs):
         """Load a ManagedBrush from disk by name, via a cache."""
@@ -751,6 +757,12 @@ class BrushManager (object):
         parent = self.selected_brush.name
         clone.brushinfo.set_string_property("parent_brush_name", parent)
         return clone
+
+    def _brush_selected_cb(self, bm, brush, brushinfo):
+        """Internal callback: User just picked a brush preset somehow (eg. from
+        a shortcut or the brush panel), so make sure a brush-dependant tool
+        (eg. Freehand, Connected Lines, etc.) is selected."""
+        self.app.doc.modes.pop_to_behaviour(gui.mode.Behavior.PAINT_BRUSH)
 
     ## Device-specific brushes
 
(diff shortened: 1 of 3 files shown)


gui/brushmanager.py
```diff
@@ -182,6 +182,14 @@ def __init__(self, stock_brushpath, user_brushpath, app):
...
         self.app.doc.input_stroke_ended += self._input_stroke_ended_cb
 
+        # In an ideal world, we would initialize the brush manager first and
+        # document second. This would allow us to place this line in the
+        # document's constructor like most callback initialization.
+        # Unfortunately, the brush manager constructor also depends on the
+        # document. (Because of the line above.)
+        # So I was forced to put this here.
+        self.brush_selected += self.app.doc._brush_selected_cb
+
```
achadwick: Here I think I'd prefer the `BrushManager` to call its own `_brush_selected_cb()`. Observing an event on the current object with a instance method of the current object seems cleaner.

(Observing far-off objects is fine too. But making far-off objects observe us feels weird.)

gui/document.py
```diff
@@ -1531,6 +1531,15 @@ def _brush_settings_changed_cb(self, *a):
...
 
+    def _brush_selected_cb(self, bm, brush, brushinfo):
+        """Internall callback: Go back to the last brush-dependant tool"""
+        # This code assumes the default mode is freehand.
+        # (Or, more accurately, it assumes the default mode is
+        # brush-dependent.)
+        brush_flags = gui.mode.Behavior.PAINT_BRUSH
+        while self.modes.top.pointer_behavior & brush_flags == 0:
+            self.modes.pop()
+
```
achadwick: What about a ModeStack API method called something like `pop_to_behaviour(self, brush_flags)`? To be called by BrushManager._brush_selected_cb().

Might be more understandable. Those `bm, brush, brushinfo` args aren't used here, they're just an artefact of on `@event` declared in some far-off observable.

ydahhrk:
Ok :)

I think I should be able to work on this later today.

(BTW: Is it normal that the tea tests failed? They don't look like Python errors so I'm assuming they were already there before I threw code in.)

achadwick:
@ydahhrk :+1: 
Thanks, much appreciated!

Tea-CI can be an interesting beast. Updates to it sometimes break 32 bit Python due to Wine issues and their image changes. Master has been adding stuff to the CI setup while your feature branch has been open - not least of which is a migration from SCons to setuptools for portability reasons. If you rebase and push to your branch, you'll get an additional real-Windows AppVeyor build check too.

Short story is, I'm not going to let a tea-ci breakage delay anything if AppVeyor and Travis both build it. Tea-CI breakages are best handled by one of us developers applying the precision adjustment tool (:hammer:) to its build matrix until we find the right combination again.

ydahhrk:
Okay, done.

Sorry; the forced push deleted your review notes so I couldn't do a final check, but I think I did everything you asked.

Notice the new import I had to add to brushmanager.py. Is that all right?

achadwick:
Thanks, this looks great! The new behaviour does feel much more comfortable.

Old reviews can be found on the "Show outdated" foldouts, btw :smile:

PR #801, opened 2017-05-15
Add Flatpak packaging

manuq:
Flatpak is a great way to distribute applications in modern GNU/Linux
distributions.

To build a flatpak package locally, follow this steps:

    cd flatpak
    flatpak remote-add --no-gpg-verify --user my-repo ~/flatpak-repo
    flatpak-builder --repo=~/flatpak-repo --force-clean my-build org.mypaint.MyPaint.json
    flatpak --user install my-repo org.mypaint.MyPaint

Then it can be run from the command line like this:

    flatpak run org.mypaint.MyPaint

A flatpak repository can be created to easily install MyPaint from a
browser with one click. Is also possible to build a bundle for sharing
with `flatpak build-bundle`.  A central place for building and hosting
Flatpak builds named Flathub is in the works. And Alexander Larsson
showed interest in adding MyPaint to it.


flatpak/mypaint-Makefile
```diff
@@ -0,0 +1,4 @@
+all:
```
Barthalion: This is mostly cosmetics, but flatpak-buillder has `simple` buildsystem since 0.9, skipping any buildsystem assumptions. [Example usage](https://github.com/flathub/org.wesnoth.Wesnoth/blob/da5aa2cb6fbc3411319665457d4a8e52a389a3fb/org.wesnoth.Wesnoth.json#L32,L46).

flatpak/numpy-Makefile
```diff
@@ -0,0 +1,4 @@
+all:
```
Barthalion: See above.

flatpak/pycairo-Makefile
```diff
@@ -0,0 +1,4 @@
+all:
```
Barthalion: See the first comment.

flatpak/org.mypaint.MyPaint.json
```diff
@@ -0,0 +1,102 @@
...
+    "runtime-version": "3.24",
+    "sdk": "org.gnome.Sdk",
+    "command": "mypaint",
+    "finish-args": [
+        "--share=ipc",
+        "--socket=x11",
+        "--socket=wayland",
+        "--filesystem=home"
+    ],
+    "cleanup": [
```
Barthalion: I guess it can be removed, if it's empty anyway?

flatpak/mypaint-Makefile
manuq: Thanks, fixed in new commit.

flatpak/numpy-Makefile
manuq: Fixed in new commit.

flatpak/pycairo-Makefile
manuq: Fixed in new commit.

flatpak/org.mypaint.MyPaint.json
manuq: Indeed. Fixed in new commit.

manuq:
@Barthalion I now added the rename-desktop-file, rename-appdata-file and rename-icon you commented in IRC. Also I added Nightly prefix. The app now looks good in my desktop. Could I ask a last review?

manuq:
@Barthalion I actually found a bug, the application code needs an icon named "mypaint", so I need a workaround to symlink all of these:

    /app/share/icons/hicolor/16x16/apps/org.mypaint.MyPaint.png
    /app/share/icons/hicolor/22x22/apps/org.mypaint.MyPaint.png
    /app/share/icons/hicolor/24x24/apps/org.mypaint.MyPaint.png
    /app/share/icons/hicolor/256x256/apps/org.mypaint.MyPaint.png
    /app/share/icons/hicolor/32x32/apps/org.mypaint.MyPaint.png
    /app/share/icons/hicolor/48x48/apps/org.mypaint.MyPaint.png
    /app/share/icons/hicolor/512x512/apps/org.mypaint.MyPaint.png

Barthalion:
You can set `copy-icon` to `true` to keep both. I will review your changes tomorrow morning.

manuq:
Adding `"copy-icon": true` to the manifest made the trick, thanks!

odysseywestra:
If you can also create a pull request about the flatpack package on the Downloads page of our website, that would help too. Once that is ready, I'll be more than happy to merge in the changes in both areas.

manuq:
@odysseywestra thanks I just sent a pull-request for the site Downloads page: https://github.com/mypaint/mypaint.github.io/pull/34

odysseywestra:
Okay I manage to install and run and looks like it works fine. One thing I would look into is to see if you can also use the dark variant of Adwaita since it is an option in our preferences.

odysseywestra:
And one of the features that MyPaint rely on is its open in external app. It should be able to see other programs or at leaset other flatpaks that can open PNG files.

odysseywestra:
Also is there an option in flatpak build scripts that allows it to install desktop files as well? It would be nice to not have to use the terminal to launch the flatpak package of MyPaint.

manuq:
@odysseywestra thanks for the review.

I tested the dark theme option in the flatpak MyPaint and it works after reopening the application.

I also tested the Open layer in external application, I used GIMP and io also worked well. I think GTK+ is using the Application Chooser portal automatically.

manuq:
@odysseywestra regarding the Open layer in external application: no, I was wrong. This is not working. I'm checking how to fix it.

odysseywestra:
Just letting you know that it is probably a better idea to to either switch this to use the master branch or create a second one pointing to master. Since MyPaint is a community project, we are slow to release. Also the issue about the external layers may have already been fixed in Master.

manuq:
@odysseywestra I fixed the flatpak. You can use the Software application to update it if your software application supports flatpak. If not, you can call `flatpak update org.mypaint.MyPaint` from command line. Could you please give it a try?

I did so by [patching the source](https://github.com/flathub/org.mypaint.MyPaint/blob/master/externalapp.patch).

Regarding working with MyPaint master, unfortunately flathub.org supports only stable releases.  I plan to provide a Nightly MyPaint for current master, the main problem is where to store it.

By the way before doing anything I checked externalapp.py in master and the content is almost the same as in the stable release.  Only changes were 1. debugging lines removed and 2. a workaround for Windows.  I will open another pull-request to propose the patch for master.  I think is great because it removes a lot of custom code that is a reimplementation of Gtk.show_uri_on_window.

odysseywestra:
Ahh, yes, using `Gtk.show_uri_on_window` work great. It even show Krita now. So, yes please, provide a PR fixing that problem. Also if you see anything else that needs to be update Gtk wise, PRs are welcome on that front.

With nightly builds, probably when Flathub comes available, we can host them there. Just keep in mind when we get bug reports we have to make sure it also affect master as well so it would be preferred. We also have the Github releases and could use the Nightly tag, but that would mean we would need to create a release repo for archival purposes. 

Lastly about the GTK theme, yes I can confirm that both light and dark variants of Adwaita theme works. I was using Arc theme and apparently the dark theme is not working, but that not your problem, but them. Or it could be the fact, that I'm doing this all on Antegros(Arch Distribution).

Overall, thank you for your doing this.

manuq:
@odysseywestra  great to know the portal is working for you too! I just posted a PR for master https://github.com/mypaint/mypaint/pull/802 . Thanks for keeping reviewing this one.

Just to clarify:

- The stable flatpak is hosted in flathub.org and the manifest lives in https://github.com/flathub/org.mypaint.MyPaint . So that is already done and usable. The only thing for stable in the PR is the flatpakref file, which is just a link to flathub.org for easy one click installation.

- This PR has turned for keeping track of the unstable / master/ nightly MyPaint flatpak manifest. I have added a README to explain this.

odysseywestra:
Okay, thank you for clarifying. So we don't really need to worry about that then.

I noticed that you are using `python setup.py managed_install --prefix /app` to install MyPaint in the json file. Managed_install is a temporary measure until we get `pip` enabled on the repository. For distribution purposes you should be using `python setup.py install --prefix /app` since you do not need to track installed files which is what `manage_install` and `manage_uninstall` do.

Lastly this my be an issue for only me on arch, but I'm gettting an error when it tries to strip the files with `eu-strip`. Here is the error it spits out on me.
```
make[1]: Entering directory '/run/build/swig/CCache'
Installing ccache-swig
Installing /app/bin/ccache-swig
/usr/bin/install -c -d /app/bin
/usr/bin/install -c -m 755 ccache-swig /app/bin/`echo ccache-swig | sed 's,x,x,'`
make[1]: Leaving directory '/run/build/swig/CCache'
Installation complete
stripping /home/albert/Projects/Github/MyPaint/mypaint/flatpak/.flatpak-builder/rofiles/rofiles-VGb44A/files/bin/swig to /home/albert/Projects/Github/MyPaint/mypaint/flatpak/.flatpak-builder/rofiles/rofiles-VGb44A/files/lib/debug/bin/swig.debug
Error: module swig: Failed to execute child process “eu-strip” (No such file or directory)

```

Barthalion:
You need to install `elfutils` as it's described in optdepends. That's under discussion between Arch and Flatpak team whether it should be a hard dependency (but apparently yes, it should).

odysseywestra:
Okay, I'm getting another error about it not being able to export MyPaint to my local repo. I'm using the instructions @manuq provided above in the first comment.

```
Exporting org.mypaint.MyPaint to repo
error: No such file or directory
```

Barthalion:
Can you check if `--verbose` gives something more informative?

manuq:
> Okay, I'm getting another error about it not being able to export MyPaint to my local repo. I'm using the instructions @manuq provided above in the first comment.

@odysseywestra note I overriden a path in my system in those commands. ~/flatpak-repo was /home/...

For quick questions like this a little chat could be better. I'm in IRC at #mypaint and #flatpak in freenode.

manuq:
I added the patch in https://github.com/mypaint/mypaint/pull/802 so you can test the Nightly flatpak with the "open layer in external app" feature enabled.

> I noticed that you are using python2 setup.py managed_install --prefix /app to install MyPaint in the json file.

@odysseywestra I just tried setup.py install and gives the following error. Instead managed_install works well.
```
Running: python2 setup.py install --prefix /app
running install
Checking .pth file support in /app/lib/mypaint/
/usr/bin/python2 -E -c pass
TEST FAILED: /app/lib/mypaint/ does NOT support .pth files
error: bad install directory or PYTHONPATH

You are attempting to install a package to a directory that is not
on PYTHONPATH and which Python does not read ".pth" files from.  The
installation directory you specified (via --install-dir, --prefix, or
the distutils default setting) was:

    /app/lib/mypaint/

and your PYTHONPATH environment variable currently contains:

    ''

Here are some of your options for correcting the problem:

* You can choose a different installation directory, i.e., one that is
  on PYTHONPATH or supports .pth files

* You can add the installation directory to the PYTHONPATH environment
  variable.  (It must then also be on PYTHONPATH whenever you run
  Python and want to use the package(s) you are installing.)

* You can set up the installation directory to support ".pth" files by
  using one of the approaches described here:

  https://pythonhosted.org/setuptools/easy_install.html#custom-installation-locations

Please make the appropriate changes for your system and try again.
```

odysseywestra:
Okay I manage to get the flatpak package to build and install. As far as the package is concern, it looks and works as intended. I would Also I would add the `/flatpak/.flatpak-builder` folder to gitnore to prevent git from seeing those file and being accidentally committed.

I suggest expanding the Readme File to include instruction to building the flatpak locally. Word in a way to build and add it to a repo it outside of the Git Repository. Here are the instructions I used for example.
```
cd ~/Projects/Github/MyPaint/mypaint/flatpak
flatpak remote-add --no-gpg-verify --user local-repo ~/Projects/Flatpaks/repo
flatpak-builder --repo=~/Projects/Flatpaks/repo --force-clean ~/Projects/Flatpaks/build/mypaint  org.mypaint.MyPaint.json
flatpak --user install local-repo org.mypaint.MyPaint
```
After doing some reading about PYTHONPATH, it is not recommended to mess with that env variable. Instead it is encourage to use `pip` or `easy_install` scripts which we do not have at the. So for the time being keep using `manage_install`, but keep in mind we will be transitioning to using those tools in the future to build MyPaint. Unless someone else has a solution to this.

Also, I'm wondering if we should just have Nightly Flatpak json file in a separate repository since it is just packaging. Technically it doesn't need the MyPaint repository, unlike our appimage scripts which use the build output that MyPaint generates. We can just place the flatpak ref files in the base of the repository if that is the case. Unless of course you can get the nightly flatpak json file to use the repository itself instead of having to re-download MyPaint Sources which seems to be a waist of bandwidth.

manuq:
> I suggest expanding the Readme File to include instruction to building the flatpak locally
OK, I added generic instructions.

> Also, I'm wondering if we should just have Nightly Flatpak json file in a separate repository
Yes indeed it can be a separate git repository. But it would be great if a flatpak build can be automatically started after a change in master. I see there is some Travis pipeline already in place. Do you think it can be used to trigger a flatpak build of unstable MyPaint? And displayed in github releases?

odysseywestra:
> Yes indeed it can be a separate git repository. But it would be great if a flatpak build can be automatically started after a change in master. I see there is some Travis pipeline already in place. Do you think it can be used to trigger a flatpak build of unstable MyPaint? And displayed in github releases?

Yeah, I think we'll keep it there for now then. Probably in the future we'll move all of the container stuff in a `/dist` folder to keep them organized and in a central place.

About Travis, I'm working on getting docker to work so that I can get updated ubuntu images. If that doesn't pan out, I'll try out https://codefresh.io since it's built around docker. So yeah, once I get to automatic deployments once I figure that out.

Lastly, I've noticed that I can't have both the flathub package and the local nightly installed at the same time. I could only launch the one or the other. The way I fixed this was adding a `-Nightly` at the end of the app-id in the json file. If there is another way to go about this, I would update the Readme file with this info to avoid confusion.

manuq:
@odysseywestra great to know you are working on a solution to have automatic builds for unstable MyPaint using Docker. I saw appimage currently uploads the file to transfer.sh .

You are right about not being able to install both the Nightly and the stable flatpak, unless the app-id differ. I think is OK to have different IDs, and appending a `Nightly` sounds good to me.

manuq:
@odysseywestra actually I was wrong: both stable and master can be installed at the same time with the same ID. In my system I have them both installed right now: the stable one from flathub and the master one from local build. To run them:

    flatpak run org.mypaint.MyPaint//master
    flatpak run org.mypaint.MyPaint//stable

odysseywestra:
Okay, what about on the user facing side? Since the desktop file takes up the name of the app-id, they are both over writing each other. Thus I'm only able to use one or the other desktop launcher. The only way I found to combat this was the `-Nightly` solution mentioned above. Unless there is a way to rename the desktop file while also retaining the ability to launch it like you can in the terminal.

Also I don't think your answer the question earlier, but would it be possible to use the MyPaint's local git repo instead of downloading it again as a tar file?

manuq:
Yes is possible to use the local checkout, just pass 'path' instead of
'url'.

I will check if is there any way to have two desktop icons.

manuq:
@odysseywestra so after discussing again with flatpak maintainer, the way to go is having different app-ids as you said. For example GIMP does org.gimp.Gimp for stable and org.gimp.GimpDevel for master.

odysseywestra:
Awesome Sauce! Thank you for clarifying. Okay, I think all that left is to have the flatpak-nightly package to use the local git repo, and I think we are good for merging your PR. ^_^

manuq:
@odysseywestra I have set path to `../` and it worked.

odysseywestra:
Yup, I can confirm that tool. Okay, I'm merging in your changes and now MyPaint now officially has a flatpak packages. If you would like to make future changes to the flatpak or MyPaint packages without PRs, request to join our @mypaint/developers team. I'm sure we would love to have you aboard. Thank you for taking the time to getting the nuances worked out. I learned a lot about flatpaks while working with you. :D

manuq:
@odysseywestra great, thanks a lot for reviewing, I've been learning a lot in the process too. Yes I would be happy to be part of the developers team.

PR #802, opened 2017-05-18
externalapp: remove OpenWithDialog and use GTK+

manuq:
Important note: this bumps the GTK+ dependency from 3.12 to 3.22 .

Since GTK+ 3.22 there is a `Gtk.show_uri_on_window()` that does the job
implemented in OpenWithDialog.

This not only removes a lot of custom code.  Is also a must to run
MyPaint in sandboxed contexts,like Flatpak or Snappy.  Because under the
hood, `Gtk.show_uri_on_window()` will automatically use the `open_uri`
portal from xdg-desktop-portal in those situations.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -47,11 +47,13 @@ installed before you can build it.
 
 - setuptools
 - pygobject
-- gtk3 (>= 3.12)
+- gtk3 (>= 3.22)
 - python (>= 2.7.4)
 - swig
 - numpy
 - pycairo (>= 1.4)
+- xdg-desktop-portal
+- xdg-desktop-portal-gtk
 
 ### Debian and derivatives
 
--- a/gui/externalapp.py
+++ b/gui/externalapp.py
@@ -33,11 +33,9 @@ import lib.xml
 
 ## UI string consts
 
-_LAUNCH_SUCCESS_MSG = _(u"Launched {app_name} to edit layer “{layer_name}”")
-_LAUNCH_FAILED_MSG = _(u"Error: failed to launch {app_name} to edit "
+_LAUNCH_SUCCESS_MSG = _(u"Launched application to edit layer “{layer_name}”")
+_LAUNCH_FAILED_MSG = _(u"Error: failed to launch application to edit "
                        u"layer “{layer_name}”")
-_LAUNCH_CANCELLED_MSG = _(u"Editing cancelled. You can still edit "
-                          u"“{layer_name}” from the Layers menu.")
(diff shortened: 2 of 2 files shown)


odysseywestra:
@achadwick I would like you to chime in on this since this will bump up the required Gtk3 version.

My consensus for compatibility sake, we may need to keep in the old code and just use this code if the user is using Gtk 3.22 or above. I don't know how feasible this is since this will introduce technical debt that we will need to keep track of.

On the bright side with using this, Krita now show up as an app available to open the layer now.

manuq:
I can provide a patch that keeps the current implementation and uses show_uri_on_window if GTK+ version >= 3.22 . I wasn't sure you were still targetting GTK+ 3.12 for master.

Note: one big benefit of distributing a Flatpak is that the app will use the desired GTK+ version (specified in "runtime-version") regardless of which version the user has installed in the system. You can check [this presentation](https://blogs.gnome.org/mclasen/2017/01/30/presentation-woes/) that explains it.

danpla:
`Gtk.show_uri_on_window()` is nothing more than a replacement for deprecated `Gtk.show_uri()`. It will simply launch the default application for a file. The user will not be able to choose between several appropriate alternatives, which is the whole point of the `OpenWithDialog`.

manuq:
> The user will not be able to choose between several appropriate alternatives

@danpla in the case of running inside a sandbox it does. So maybe this behaviour is only when running in a sandbox.
captura de pantalla de 2017-05-19 16-59-49

danpla:
@manuq Maybe it's desktop-specific. Does the old `Gtk.show_uri()` show the dialog? On my XFCE, it always calls the default app.

manuq:
@danpla I don't think is desktop-specific, but depending if the application context is a sandbox or not. If in a sandbox, it will use the open_uri portal like explained [here](https://blogs.gnome.org/mclasen/2016/07/08/portals-using-gtk-in-a-flatpak/)

danpla:
@manuq Ah, got it. But, as you mentioned, it will depend on sandboxing provided by Flatpak (or similar thing), while most users install apps (including MyPaint) via package managers. In this case (and on Windows), there will be no dialog.

manuq:
@danpla yes you are right. So I will have to add a check for sandboxing. Thanks for noting it.

manuq:
I am closing this for now since GTK+ only provides this feature for sandboxed applications and I will keep patching the code while sandboxing MyPaint  https://github.com/mypaint/mypaint/pull/801 .

PR #812, opened 2017-06-10
Curve points

briend:
The limit of 8 points is a bit too restrictive, preventing
you from making stair-stepped or other interesting patterns
that are especially useful for stroke, anti-art offsets, gridmap, etc.
This patch requires the corresponding patch to libmypaint.

I've added 64 points, and made the grid 10x10 instead of 4x4.
Also, the points snap on 0.05 instead of 0.25
screenshot 2017-06-09 22 17 00

### DIFF
--- a/gui/curve.py
+++ b/gui/curve.py
@@ -14,14 +14,14 @@ logger = logging.getLogger(__name__)
 
 from gi.repository import Gtk, Gdk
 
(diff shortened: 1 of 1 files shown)


achadwick:
No problem with these changes. I've often found the nonlin points graph somewhat restrictive.

PR #814, opened 2017-06-16
inputs: add barrel-rotation support (Art Pen, etc)

briend:
initial barrel-rotation support. -180 to +180.  Tested with Wacom
Art Pen. Requires update libmypaint to support new input.  It seems
many pens will provide bogus rotation values so a method to disable
rotation seems to be needed.  This is the GDK Wheel device which
can vary quite a bit from wheels to hockey pucks/dials, etc.  Needs updated libmypaint.
screenshot from 2017-06-15 20-47-42

### DIFF
--- a/gui/drawutils.py
+++ b/gui/drawutils.py
@@ -40,14 +40,14 @@ from lib.pixbufsurface import render_as_pixbuf
 ## Module constants
 
 _BRUSH_PREVIEW_POINTS = [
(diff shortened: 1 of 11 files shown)


briend:
My plain wacom stylus (non-rotating) still reports a rotation of -900 via xinput.  Anyone have any ideas on how to handle that?  A preference to turn on/off rotation would be annoying in the case where you want to switch to the Art Pen and then go back to a normal pen.  Maybe there is some other identification we can use. . .

PR #825, opened 2017-06-23
viewzoom brush tweaks. Missed a few offset_by_speed.

briend:
Very few brushes used this setting.


odysseywestra:
@briend for future reference, if you have a commit you are confident in, go ahead and directly push to master. If a bug happens to pop up, you can either revert the commit or fix the bug.

briend:
Ah, that'll keep the commit log a lot cleaner I bet.  Thanks!  I hope that's the end of this :-)

PR #836, opened 2017-07-12
ColorPicker HCY:  Pick H, C, or Y from canvas

### LINKED ISSUE #835: Pick H,C and/or Y from canvas
Demo here:
https://www.youtube.com/watch?v=ZLt8DrzDUlc&lc=z12bezzhnvayfpimx23ahhbydqizyr4do
I made a branch here
https://github.com/briend/mypaint/tree/pickHCY

The idea is, say you have some colors on your canvas.  You want a yellow color that sorta matches (coordinates with) a blue color on the canvas.  You can just pick any yellow color from palette and then alt+r and/or shift+r to modify the yellow color to have the same chroma and/or luma as the blue on the canvas.  Kinda cheating!

Or maybe you already have a scene on your canvas and you want to paint a section that has diminished chromacity and a brighter luma (like something in the distance).  So you might be able to paint one element in the distance and use that as the springboard/model to pull foreground colors into the background space by grabbing foreground colors (R) and then plucking the C and/or Y (alt+R, shift+R) from your distant model.

Or maybe your paint brush already has a color that you like, but you want to change the hue of it to match something else on your painting, but preserve the chroma and luma.  Just press Ctrl+R now instead of trying to rotate the hue wheel around to the approximate color.

Or maybe this doesn't make any sense at all :-D

Making color ramps like this is really really simple and flexible by just using the hot keys to grab the luma and chroma of the previous swatch to paint the next one.  Previously it would be just about impossible.
screenshot from 2017-07-12 18-48-57

Here's a youtube clip demonstrating a bit:
https://www.youtube.com/watch?v=_YySR2PQJ30

### LINKED ISSUE #838: Preserve brush Hue when picking achromatic colors from canvas
Currently if you "color pick" anything on that canvas that lacks color, the hue gets reset to "red".  This can be problematic if you are using a brush that has dynamic colors adjustments, or if you use the keyboard shortcuts for lighten/darken or increase/decrease saturation.  Or if you were planning to re-adjust your color with the sliders for saturation, and brightness.  The color wheels and sliders all preserve the Hue even for achromatic colors, so I think it makes sense for the color picker to do the same.

I can see why it behaves that way-- there's no hue information in an RGB value like 1,1,1, so hue=0 is logical.  Except, we DO know what the previous brush color was, and it probably makes more sense to just keep that hue instead.  It should be pretty easy to tweak the colorpicker code so that it compares the new pick_color with the old brush color and, in an achromatic case, set the new brush color hue to the old brush color's hue.

briend:
Hold down control, shift, or alt plus 'r' to modify brush color in the H, C, or Y dimensions, respectively.
- [x] Colors picked in HCY
- [x] Avoid reset to Red for achromatic picks
- [x] ~~Make color space a preference?~~  HCY really is great, IMO
- [x] ~~Better keyboard shortcuts?~~ Let user decide
Discussion here:
https://github.com/mypaint/mypaint/issues/835 
closes #835  
closes #838


briend:
Could someone help me add arguments to colorpicker.py?  This PR I have just copied the file 3 times for each of H, C, and Y and added a few lines to switch the RGB to HCY.  I don't really know how to go about getting it to take an argument, any pointers would be appreciated!

briend:
Looking quite a bit better.  I had to check for Chroma and Luma <0.01 so as to avoid the rounding errors that would flip the hue back to red.  Please give it a try.  The hue should never flip to red with any of the color picking modes now.  I probably need to tidy up some more things and rename variables a bit to make them more sensible.

briend:
I lied, there was another corner case still flipping hue to red.  I refactored so it might be readable code now.  Now I appreciate why no other program seems to do this.  You can't retain hue information when converting from hsv-->rgb-->hcy and back again unless you constrain your C to >0.01 and your 0.01 < Y < 0.99.  Really fun!

gui/resources.xml
```diff
@@ -1779,39 +1779,33 @@ Vocabulary
       </child>
       <child>
         <object class="GtkRadioAction" id="ColorPickModeH">
-          <property name="label" translatable="yes" context="Menu→Edit→Current Mode (labels)">Pick Color Hue</property>
-          <property name="short-label" translatable="yes" context="Toolbar (short labels)">Pick ColH.</property>
-          <property name="tooltip" translatable="yes" context="Toolbar (tooltips)">Pick Color: set the painting color Hue from screen pixels.</property>
           <property name="icon-name">mypaint-pick-color-h-symbolic</property>
           <property name="group">SymmetryEditMode</property>
           <property name="value">11</property>
+          <!-- NO TOOLTIP OR LABEL. -->
```
briend: Should every action have a menu item?  Seems like 4 colorpick modes in the menu is overkill but not sure

briend:
The only problem now is that if you release the modifier *before* the "r" key, you might trigger a normal pick mode if the cursor was moving.  Plus, this normal pick mode popup gets "stuck" until you press "r" again.

PR #839, opened 2017-07-14
ColorPicker: Preserve brush Hue when picking achromatic colors.

### LINKED ISSUE #838: Preserve brush Hue when picking achromatic colors from canvas

briend:
This allows you to continue to use increase/decrease saturation
adjusters and/or saturation brush dynamics even after picking a
grey/black/white color from canvas.  Normally this would flip
the Hue to red, which is likely to be unexpected.
closes #838

### DIFF
--- a/gui/colorpicker.py
+++ b/gui/colorpicker.py
@@ -84,6 +84,15 @@ class ColorPickMode (gui.mode.OneshotDragMode):
             t, x, y = doc.get_last_event_info(tdw)
             if None not in (x, y):
                 color = tdw.pick_color(x, y)
+                colortest = color.get_hsv()
+                
+                brushcolor = self._get_app_brush_color()
+                brushcolor = brushcolor.get_hsv()
+                
+                if (colortest[1] == 0) or (colortest[2] == 0):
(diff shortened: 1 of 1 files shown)


briend:
Should I create a function that takes x,y and sets the brush?

briend:
closing as I will just include this in the HCY color picker thingy

PR #853, opened 2017-08-24
CAM16, Pigment Color adjusters and Swatch Overlay, etc [WIP]

briend:
Discussion [here](https://community.mypaint.org/t/hcy-adjuster-keys-allow-tints-shades-tones/894)
I'm investigating CIECAM as a color adjuster method.  It provides many more advantages compared to HCY or HSV.

- [x] CAM16 gui sliders
- [x] Color Adjuster keyboard shortcuts configurable via pref.  Use HSV, HCY, CAM16, or Pigment models
- [x] Limit color purity to an adjustable level to help enforce a limited palette
- [x] Adjustable illuminants: A, D65, D55, D50,  and Custom XYZ (via colorpicker) 
- [x] Make adjuster step size relative to how frequent you press an adjuster key.  Quick=larger step. Slow=smaller step. 
- [x] Make color adjustment step size static or dynamic, both with adjustable step sizes
- [x] Implement a floating overlay similar to ColorPicker that shows you your new adjustments as a borderless swatch on top of your painting
- [x] Make the swatch overlay a set of two preferences: on/off during stroke, and on/off before stroke.
- [x] Added CIECAM UIColor class
- [x] Added Pigment UIcolor class- spectral upsampling w/ subtractive paint-like properties
- [x] Added LinearRGB UIColor class
- [x] Set a maximum rate for color pickers/adjusters to avoid cpu spike (10hz)
- [x] Constrain to sRGB gamut and display limits (0-1) to avoid hue shifts
- [x] Store the illuminant and entire ciecam state within the strokemap for stroke-picks
- [x] Use kmeans dominant color for color picker.  Adjust sample size w/ zoom level.
- [x] Map a key to apply a canvas color as an illuminant to the current brush color
- [x] Interactive Blend mode, ctrl+alt+R will allow blending brush color with canvas color by moving cursor back and forth; uses spectral Pigment model.  Thank you @0ion9 for implanting this interface idea in my brain a long time ago!
- [x] New gradient modes, CIECAM, Opaque Paint, Crayon, Watercolor Glaze
- [x] Fully customizable slider preferences, show/hide any particular slider via preferences.


achadwick:
> I would like to also constrain the color to the gamut mask:

Needs the stuff in the HCY colour selector that does the constraints with user-defined polygons to be moved into the main gui.brushcolor.BrushColorManager instance.

The polygons should ultimately be moved to the model, but that requires saving arbitrary JSON to ORA (working on it...)

briend:
Eek, maybe I'll leave that off this PR then, and revisit that someday :-)

briend:
So, this is getting pretty cool and hopefully useful ;-).  I imported colorspacious and added preferences to allow you to select CIECAM02 axes as well as light sources (maybe I'll add a custom XYZ light source too).  SO, if you choose a warm light source and desaturate a red color you get a totally different result compared to a cool light source.  @achadwick what do you think?  Can we add python-colorspacious to the build environment? :-D
screenshot from 2018-01-27 19-31-48

briend:
I've got a UIColor mostly working w/ CIECAM:  Gradients seem much nicer imo:

top to bottom; rgb, hsv, hcy, ciecam
screenshot from 2018-01-30 20-32-44
screenshot from 2018-01-30 20-30-57

briend:
Here's a quick demo of changing the illuminant.  The center of each flower is the grabbed illuminant and the pedals are the result of rotating the hue "around" this illuminant (with a bit of chroma increase).  It's a pretty quick process to build out some color schemes:
ciecam-demo

briend:
Changed some logic so that when you select colors from gui interfaces, palette, etc, they are "white-balanced" for the selected illuminant.  This can be really handy if you have a given palette that you want to represent under different lighting conditions.  Here is the same palette under a variety of illuminants:
illumtest

briend:
I submitted colorspacious to MINGW-packages and it was merged, so the next commit I do should pass doctests and build properly.  I didn't understand styrene at all and was trying to pip install colorspacious :-D

briend:
Darn.  Colorspacious is definitely available in MSYS2 now, but I guess the build environment doesn't update every time?

briend:
Sorry hound

ValZapod:
Cannot you just use clang format to format stuff? You are never supposed to do it by hand...

Issue #392, opened 2015-08-03
Move codebase to Python 3


achadwick:
Python 3 has been out since 2008. It is high time the MyPaint codebase was updated to use it.

(Simple description, lots of work, might run into some scons issues...!)

QuLogic:
I probably should have asked this earlier, but are you planning on supporting both 2 and 3 simultaneously?

achadwick:
Hmm, good question, given that we build the python extension. Can scons reliably generate either, on all platforms? I'm not sure it can.

It may be worth moving to distutils/setuptools first.

odysseywestra:
What left thats need to be ported to python3?

jplloyd:
There may be some minor issue lurking still, but otherwise everything is now compatible with Python 3 (excluding the icon export script - easily fixed).

PR #880, opened 2017-12-04
Port SCons files to python3

bmwiedemann:
using different notation for ocatal numbers,
using brackets for print()
using 'in' instead of dropped has_key

all changes are backward compatible (tested with python-2.6.9)

helps with #392 

Background: python2 will be discontinued in 2020 thus we are pushing it out of openSUSE Tumbleweed now, which causes build failures with the mypaint package now.

### DIFF
--- a/SConscript
+++ b/SConscript
@@ -14,9 +14,9 @@ mypaintlib = SConscript('lib/SConscript')
 languages = SConscript('po/SConscript')
 
 try:
-    new_umask = 022
+    new_umask = 0o22
     old_umask = os.umask(new_umask)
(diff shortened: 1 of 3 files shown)


achadwick:
Hi there. No objection to this simple Python3 compat patch for a thing we don't recommend or test, but please be aware that we are planning to get rid of SCons entirely due to its woeful Python3 and Windows/MSYS2 support. Do you think you could work on updating the setup.py script and the main Python code to be Python3 compatible as well?

If you are building from master, your build should use setuptools and setup.py.

I think we're in a good state to go Python3 for the main program code fairly soon.

PR #891, opened 2018-01-22
HCY Luma correction for Rec 709 sRGB

### LINKED ISSUE #890: Use Rec 709 luminance weights and linear for HCY tools
I'm creating a patch here:
https://github.com/briend/mypaint/tree/LumaFix1

Some background from Bruce Lindbloom:
http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html

> Perhaps you've learned that you can compute the luminance of an RGB color by taking 30% of its red component plus 59% of its green component plus 11% of its blue component. These weightings are often expressed in three-digit precision as 29.9% red, 58.7% green and 11.4% blue. Did you ever wonder where these weightings came from?

>You can find them in the above table as the relative Y values for red, green and blue for the NTSC color model. The more precise weightings are 29.8839% red, 58.6811% green and 11.4350% blue. But it should also be obvious that the real RGB weightings depend upon the color system in use. So the "standard" weightings are incorrect for other RGB systems like sRGB or Adobe RGB (1998).

>Another relevant fact is that these weightings must be made in a linear RGB space, that is, after the gamma companding function has been removed. It is very common to see the weightings applied bluntly to the companded RGB values, which is wrong.

On the left is unpatched mypaint and an attempt to "luminance match (Y)"  (using alt-R or pick luma- Luma is the wrong word here) the yellow color with the other colors.  As you can see, it is pretty unsatisfactory.  On the right is a patched mypaint with the sRGB luminance weights and the functions to convert to linear beforehand
lumafix

Here we can verify correctness by loading into Krita and converting the image to linear, then performing a desaturate- as you can see the colors disappear into each other because the luminosity is the same.
lumafix2

There are several things affected, the HCY circle, guis, and even the HCY palette "fill gap" option.  The bottom row of the palette in each screenshot was done with an HCY Fill Gap and the results are quite a bit different (better imo).

It would be really nice to get all the luminance  stuff to match and be "correct".  I know W3C has this standard here that we were following:
https://www.w3.org/TR/compositing-1/#blendingnonseparable

I can't really explain why the W3C is wrong, but it is :-).  The problem is if we fix the blending modes it will change the way existing images look.  At least with this current patch it should only affect the visuals of the interface and not break any old files.

briend:
helper functions for sRGB<->Linear added and luma weights corrected

Please see #890 for details/discussion

closes #890

### DIFF
--- a/gui/colors/sliders.py
+++ b/gui/colors/sliders.py
@@ -285,12 +285,12 @@ class HCYLumaSlider (SliderColorAdjuster):
 
     def get_color_for_bar_amount(self, amt):
         col = HCYColor(color=self.get_managed_color())
-        col.y = amt
+        col.y = amt**2.4
         return col
 
     def get_bar_amount_for_color(self, col):
         col = HCYColor(color=col)
-        return col.y
+        return col.y**(1/2.4)
 
     def get_background_validity(self):
         col = HCYColor(color=self.get_managed_color())
--- a/lib/color.py
+++ b/lib/color.py
@@ -30,7 +30,7 @@ import struct
 import gi
 from gi.repository import GdkPixbuf
 
-from lib.helpers import clamp
+from lib.helpers import clamp, srgb_to_linear, linear_to_srgb
 
 
 ## Lightweight color objects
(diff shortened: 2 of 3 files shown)


briend:
closing this.... going to suggest something else as the problem is much bigger than I thought (as usual)

Issue #791, opened 2017-03-17
v1.2.1: Cannot start on Fedora 25 Wayland


dyskette:
#### Description of the problem
mypaint cannot start under wayland, just showing a TypeError in the terminal. 

- It works fine under Xorg.

#### Basic system details

MyPaint version: 1.2.1 stable
Operating system: Fedora 25
Desktop environment: GNOME Wayland

#### Steps to reproduce

1. Install it from here: [32 bits](https://copr-be.cloud.fedoraproject.org/results/dyskette/tests/fedora-25-i386/00528144-mypaint/) or [64 bit](https://copr-be.cloud.fedoraproject.org/results/dyskette/tests/fedora-25-x86_64/00528144-mypaint/)
2. Open mypaint in GNOME wayland session

#### Backtraces or error messages

```python
INFO: mypaint: Installation layout: conventional POSIX-like structure with prefix u'/usr'
INFO: lib.i18n: POSIX: LANG='es_MX.UTF-8'
INFO: lib.i18n: POSIX: LANGUAGE=None
Gdk-Message: Unable to load exchange from the cursor theme
Traceback (most recent call last):
  File "/usr/bin/mypaint", line 464, in <module>
    main.main(datapath, iconspath, old_confpath, version=version)
  File "/usr/share/mypaint/gui/main.py", line 95, in main
    from gui import application
  File "/usr/share/mypaint/gui/application.py", line 54, in <module>
    import filehandling
  File "/usr/share/mypaint/gui/filehandling.py", line 31, in <module>
    import drawwindow
  File "/usr/share/mypaint/gui/drawwindow.py", line 39, in <module>
    import quickchoice
  File "/usr/share/mypaint/gui/quickchoice.py", line 25, in <module>
    import gui.colortools
  File "/usr/share/mypaint/gui/colortools.py", line 20, in <module>
    from colors.hcywheel import HCYAdjusterPage
  File "/usr/share/mypaint/gui/colors/hcywheel.py", line 379, in <module>
    class HCYMaskEditorWheel (HCYHueChromaWheel):
  File "/usr/share/mypaint/gui/colors/hcywheel.py", line 397, in HCYMaskEditorWheel
    __rotate_cursor = gdk.Cursor(gdk.EXCHANGE)
  File "/usr/lib64/python2.7/site-packages/gi/overrides/Gdk.py", line 329, in __new__
    return cls.new(*args, **kwds)
TypeError: constructor returned NULL
```

### Solution
[...]

PR #911, opened 2018-03-25
Load cursors in 'enter' handler

avsej:
It could also use `Gdk.Cursor.new_for_display`, but I found it handy to use `self.doc.app.cursor.get_action_cursor`

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1555277

### DIFF
--- a/gui/mode.py
+++ b/gui/mode.py
@@ -24,6 +24,8 @@ from lib.observable import event
 from lib.pycompat import add_metaclass
 from lib.pycompat import unicode
 
+import gui.cursor
+
 logger = logging.getLogger(__name__)
 
 ## Module constants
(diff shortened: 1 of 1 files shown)


achadwick:
Ctrl+F "wayland". Yep, not surprised.  I guess this relates to #791 too.

Thanks for the contribution. I think I like the approach. Let's give that a try.

avsej:
Is it going to be new release soon? Is there some roadmap or schedule?

achadwick:
Roadmap uncertain. If 1.3.0 is soon, it will be a Py2 release unless there are pressing reasons for it not to be.

Anyway, your patch tests out OK for drags and singleclicks. Thanks!

PR #917, opened 2018-04-02
Add right-click and dragging functionallity to the brush icon in the bottom right, and the popup brush list.

JornVernee:
I was having some trouble making the adding brushes to groups work, and so were others: https://community.mypaint.org/t/add-brushes-to-favorites/499

Although things worked as expected for the dockable brush group tool window, they did not for the brush selection popup that appears when clicking the brush icon in the bottom right.

I looked into the code and it seems like the latter uses a `BrushList`, a subclass of `PixBufList`, but the former just uses `PixBufList`, which doesn't have the right-click and dragging functionality that's needed to add a brush to a group.

So this PR simply changes the `PixBufList` used by the popup to a `BrushList`, which already had the needed functionality.

I've done some quick manual testing and it seems to just work. I can now drag a brush from the popup to a docked brush group window. I can also right click a brush icon to get a context menu (and add to favorites)

Furthermore, I added a right click menu, and drag functionality to the brush icon in the bottom right. It works similar to the brush icons found in a brush group window, except there's no remove brush/clone brush functionality in the right-click menu. I had to modify the BrushList to know the new widget I added as a drag source, not sure if more of em need updating...

---

There is one unintended change that this brings with it, which is that the first popup (but only the first) that's opened is miss-aligned with the footer (instead it seems to align with the bottom of the app window). The alignment doesn't happen until after the popup is drawn for the first time.  I've done some digging but I'm not sure how that could be changed though. Perhaps some update-position-y type of call somewhere in `BrushChooserPopup` or `ChooserPopup` will do the trick. Any ways, the added functionality seemed like a win, so I'm making this PR.

---

I also added the `out` folder to the gitnore, since not all the binary files in there were ignored, hope that's okay to put on here too?

### DIFF
--- a/glade/mypaint_widgets.xml
+++ b/glade/mypaint_widgets.xml
@@ -300,6 +300,16 @@
             optional-default="True" />
       </properties>
     </glade-widget-class>
+    <glade-widget-class name="MyPaintBrushWidget"
+        title="Brush widget" parent="GtkDrawingArea"
+        generic-name="brushwidget">
+      <properties>
+        <property id="width-request" default="48" optional="True"
+            optional-default="True" />
+        <property id="height-request" default="48" optional="True"
+            optional-default="True" />
+      </properties>
+    </glade-widget-class>
 
   </glade-widget-classes>
 
@@ -317,6 +327,7 @@
     <glade-widget-class-ref name="AccelMapEditor"/>
     <glade-widget-class-ref name="MyPaintDeviceSettingsEditor"/>
     <glade-widget-class-ref name="MyPaintBrushColorIndicator"/>
+    <glade-widget-class-ref name="MyPaintBrushWidget"/>
   </glade-widget-group>
 
 </glade-catalog>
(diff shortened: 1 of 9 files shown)


JornVernee:
Not sure what hound still wants here. Any way to run the linter locally? This is getting quite tedious...

JornVernee:
Looks like these are some delayed comments...

briend:
Thanks Jorn, can you rebase and squash your commits to just one or two (however many make sense)?

JornVernee:
@briend I hope that worked, I've only had to rebase once before.

odysseywestra:
@JornVernee We're about to do a feature freeze soon and would like to know if can get your PR updated so we can review and test it before merging it in.

JornVernee:
@odysseywestra Sorry, but I don't have any time I could spend on picking this up again for at least the next 2 months (probably more though...), and at this point I don't think I really want to either (I've not really used MyPaint since then, and I'd basically have to get into it again from scratch).

jplloyd:
Notes for the future: The right-click functionality does not work (menu does not show up, at least w. Gtk 3.18), but the rest works fine.

The Travis CI only fails due to an `from x import y` in `gui/quickchoice.py` that Python 3 can't handle (insert `.` before module to fix).

I don't want to trigger the CI and waste another 230MB in Appveyor right now, so we'll leave it like this for the time being.

JornVernee:
Closing old PR

PR #937, opened 2018-06-25
WIP: Add simple print operation

### LINKED ISSUE #888: A very simple print function would be really nice
Please include a simple print function (again).

The [faq](https://github.com/mypaint/mypaint/wiki/FAQ-Printing) points out that there is no printing function planned for MyPaint right now. The reasons pointed out there are fine, but they cite the specific needs of power users.

However, MyPaint targets novice users or those who want a simple, distraction-free graphics program. The user in my home is a child. My child wants to draw stuff and then easily print it so that she can show her work to her friends.

A power user can still export the file and do post-processing or colour management with external software, things that are more complicated and beyond the capability or interests of a novice user.

Thanks.

hzulla:
Hello,

this implements a very simple print function and closes #888.

There are still some issues with this, therefore I'm asking for comments.

The print settings are being reset each time the user creates a new or loads an existing document. For the first printing, the documents dimensions and resolution are compared to those of the printer's default document size.

If the document fits onto a single printer's page, it is scaled at the default 100%, even if that results in a small document printed on an mostly empty page.

If the document dimensions and resolution are too large, printing is scaled down accordingly to fit a single page.

If the user decides to use a different scaling factor or printer document size, the document will be tiled on as many printer pages as needed for the user's desired settings.

The print settings are then stored for the document and will be reused for the next call to the print function. If the user has made the document larger by then, the printing will be tiled accordingly.

Here are the issues where I request help/feedback.

- The accelerator `Ctrl`+`P` doesn't work as expected, it will activate the `TogglePreviewTool` instead and I haven't found out why.
- My code first writes a temporary PNG file, uses that for printing and then deletes it later. Obviously, it should be possible to use the document directly for printing without a temporary image file, but my understanding of mypaint's cairo/gtk code is very limited and bit of help on how to make the document draw directly on the print page's cairo context  that would be welcome.
- The default DPI setting of a mypaint document is 72, which results in rather big prints. But that may be desirable for younger users, so it isn't much of an issue.
- Should the document always be printed on the center of the printer page?

### DIFF
--- a/gui/filehandling.py
+++ b/gui/filehandling.py
@@ -22,7 +22,10 @@ import sys
 import logging
 from collections import OrderedDict
 import time
+import math
+import tempfile
 
+import cairo
 from gi.repository import Gtk
 from gi.repository import Pango
 
@@ -529,6 +532,7 @@ class FileHandler (object):
             'jpeg-90%': _SaveFormat.JPEG,
             'png-solid': _SaveFormat.PNG_SOLID,
         }
+        self.print_settings = None
 
     def _update_recent_items(self):
         """Updates self._recent_items from the GTK RecentManager.
@@ -802,6 +806,7 @@ class FileHandler (object):
         self.doc.reset_background()
         self.doc.model.clear()
         self.filename = None
+        self.print_settings = None
         self._update_recent_items()
         self.app.doc.reset_view(True, True, True)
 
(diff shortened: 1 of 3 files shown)


hzulla:
> continuous-integration/appveyor/pr — AppVeyor build failed

Please advise. It appears to be unrelated to my patch?

PR #952, opened 2018-12-07
Brushes data

briend:
few tweaks of @Jehan patch for windows build.


briend:
lets see how badly this breaks

PR #1016, opened 2019-07-16
Windows build brush paths change + misc fixes

jplloyd:
This fixes the 32bit windows bundles/installers and properly ignores the generated config file lib/config.py

~~Makes adding the relative brushdir path in the windows bundle part of the makepkg packaging step.
Overwrites the previous value configured in build_config for these bundles-~~

~~Alternative approach here is to add an option to the BuildConfig command for taking a static string instead of fetching it via pkg-config (and pass the same generated path that is used in this changeset).
More things to change, but perhaps better?~~

Edit: Now adds a new option in BuildConfig that takes the relative path, so that lib/config.py is only written to from a single place.

PR also includes some minor code qual fixes and a py3-compat fix.


jplloyd:
I've verified that both the 32bit and 64bit build artifacts work now, merging.

PR #1019, opened 2019-07-24
Configurable locale

jplloyd:
This branch adds support for user-configurable locale (without explicitly setting the LANGUAGE environment variable before running). Given the completeness and consistency of some of the translations, this is undoubtedly needed.

Do not merge this. I have not tested it on all platforms yet.

Questionable things to be found here: 
* ~~supported locales are determined only on the existence of *.po files, not on their respective completeness.~~
**There is now a build option that limits included translations based on a given completion percentage. This affects both which locales are installed and which languages can be selected in the preferences.**

* The list of languages needs to be manually updated whenever a new language is added.
**Rather a dictionary mapping locales to language names - but the test suite now verifies that names exist for all languages**

* The locale is set by overriding LANGUAGE in os.environ before running the gettext initialization
**Still the case... could probably be done better.**


jplloyd:
Tested and confirmed to work with: 
* Ubuntu 16.04 (64bit)
* Windows 7 (32bit) - standalone
* Windows 10 (64bit) - standalone*

I do think this functionality is needed, but would not mind input on how it could be done better.

_*On a side note, standalone still uses non-standalone user settings, so it's not fully portable at the moment._

jplloyd:
A note about the status of this PR: I'm holding off on rebasing and merging this until the other translation-related tasks are completed.

The appveyor build failure is unrelated to this code (it was caused by an update to the appveyor build environment).

PR #1031, opened 2019-09-06
ellipsis, added

leela52452:
[source](https://hosted.weblate.org/projects/mypaint/mypaint/source/review/?type=check%3Aellipsis)

### DIFF
--- a/gui/brushselectionwindow.py
+++ b/gui/brushselectionwindow.py
@@ -531,13 +531,13 @@ class BrushGroupsMenu (Gtk.Menu):
         # Static items
         item = Gtk.SeparatorMenuItem()
         self.append(item)
-        item = Gtk.MenuItem(C_("brush groups menu", "New Group..."))
+        item = Gtk.MenuItem(C_("brush groups menu", "New Group…"))
         item.connect("activate", self._new_brush_group_cb)
         self.append(item)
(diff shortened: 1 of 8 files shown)


jplloyd:
Thanks for the work, but there are a couple of problems.

1. You may notice that the nose tests fail when running on python 2.7. This is because an ellipsis is not a valid ascii character, and there is no explicit encoding declared.  The tests can be made to pass by adding an explicit encoding at the beginning of all files with non-ascii characters:
```
# -*- coding: utf-8 -*-
```

2. We would also need to update the msgid strings in the translation files, and I'd rather do this after pushing some outstanding changes related to older unmerged "new" strings (mostly minor changes). I will push those during the weekend.

3. You don't have to make one commit for each file. This should really be a single commit with a message like
```
Replace ... with … in user-facing strings

Use unicode ellipsis glyph instead of ... in the translateable strings
to make it less confusing for translators.
```

4. Just a general note: when making a pull request, do tell why the change is needed when it might not be obvious. I was guessing about translator convenience/confusion, is this the case?


If you add the encodings and rebase-squash into a single commit, with an appropriate message, I will merge it when I'm doing the other changes on Saturday or Sunday (CET).

jplloyd:
Changes are now part of #1032 , so this will be included when that is merged.

PR #1032, opened 2019-09-09
String updates

jplloyd:
Merges all of the previously unmerged strings - with a few fixes before being merged - into the translation files.

Also incorporates the patch from @leela52452 (modified slightly by me) to replace the occurrences of triple dots in the code strings (non-xml) with ellipsis glyphs. In order to avoid unnecessary work I have updated the msgid and (most) msgstr entries accordingly - just merging as usual would have marked all those strings as fuzzy, requiring manual intervention to be used by gettext,


jplloyd:
Due to changes to the user-facing strings not being merged for over 4 years, this change-set will mark up to ~170 (~13.5%) translated strings (per language) as fuzzy.

Most "fuzzy" marks are probably needed, but a number of them have causes that are simple enough to be potentially manageable by scripting - marks caused by variable renaming, changes from c-style formatting to python brace formatting, added contexts (and nothing else), etc.

It feels bad to create unnecessary work for our translators, but I think the safest course of action is to just bite the bullet and push this update. Any opinions on this are appreciated. If there's no disagreement I will leave this up until Wednesday before merging.

Issue #986, opened 2019-03-14
[Artwork] New Icon Proposal


jimmac:
I'd like to propose an update of style for Mypaint app icon. While the current icon is a great piece of 3D art on its own, the world has moved on from a highly detailed realistic renderings of objects as app icons in all contemporary operating systems.

For more background on the issue feel free to read my two blog posts on app icon design [1][2]

With the new gudelines in mind and the effort to keep the current metaphor, here's my intial design proposal. 

![Mypaint](https://user-images.githubusercontent.com/14821/54353913-89d34680-4655-11e9-85d1-9f3b38ca7c96.png)

![mypain-fullcolor](https://user-images.githubusercontent.com/14821/54353961-ac655f80-4655-11e9-9430-10c7015d8551.png)

Apart from the fullcolor icon, provided as a scalable vector graphic file (the only 'size' it comes from now on), I've updated the symbolic variant of teh app icon to be used mainly by the high contrast/accessible theme.

I usually try to make it effortless for maintainers to incorporate the icon and provide patches[3], but sadly have trouble building master. Thus I am attaching the SVGs only and hope you can find the time to update the build system side of things.

[1] http://jimmac.musichall.cz/blog/2018-07-18-detail-considered-harmful/
[2] http://jimmac.musichall.cz/blog/2019-01-23-the-big-app-icon-redesign/
[3] https://gitlab.gnome.org/GNOME/Initiatives/issues/2

jimmac:
Scalabel and symbolic variants of teh app icon as a zipped archive (contains 2 svg files)
[mypaint-icon.zip](https://github.com/mypaint/mypaint/files/2966173/mypaint-icon.zip)

manuq:
The proposed icon looks great, @jimmac . I hope it lands soon!

tjhietala:
While I like the new Gnome icon style in general, this icon needs improvement, IMO. It doesn't pass the squint test: the shaft of the brush is hard to see and the head of the brush seems more like a giant blueberry. The overall shape of the brush doesn't "pop out". (The initial sketch is much better than the final result in this regard.)

I'm afraid I don't have the vector graphics skills to improve it myself but hope you take this as constructive criticism.

odysseywestra:
@jimmac are you still willing to work on getting this icon improved?

jplloyd:
Original
[...]

PR #1051, opened 2020-01-20
icon: new app icon

jimmac:
- follows the GNOME HIG 3.30
  see https://gitlab.gnome.org/GNOME/Initiatives/issues/2
- extra assets like the symbolic icon and a nightly variant are
  provided but not used

Fixes https://github.com/mypaint/mypaint/issues/986


jplloyd:
Apart from the build not passing, this commit deletes a bunch of icons used for the color selectors. For obvious reasons, these are not symbolic. Whether they should be installed where they currently are is perhaps questionable, but they should not be deleted outright.

jimmac:
The action context deletion was not on purpose.

jplloyd:
I removed the remaining old app icon for 22x22, and fixed up a test that relied on the scalable icon having dimensions >= 256x256. That flake8 warning can be safely ignored (and will be explicitly ignored soon enough).

PR #1053, opened 2020-02-09
Enable rpath on BSD systems

myfreeweb:
rpath doesn't seem necessary for system installation but could be useful still.


jplloyd:
If this becomes relevant for more platforms we'll put the conditionals in a function, but for two platforms this is fine for now. Thanks!

PR #1058, opened 2020-02-19
exec() needs 1st arg to be str, not a file object.

badshah400:


jplloyd:
Good catch, I really should go through all files to check for lingering non-Py3-compatible stuff.

Couple of things to fix. 

**1.**  
The fix is wrong. `.open()` should be `.read()`.

**2.**  
For the commit message we try to consistently use imperative mood - meaning that the commit message should be written like a command; "do x using y". Since this is a python 3 change, we would also use a category prefix to indicate the domain or reason for the change. In this case the message might be:

`Py3: use a string as the 1st argument to exec`

Since this is a one-liner that does not touch anything used by the ci, you can* add `[skip ci]` to the end of the message (only works if it is on the first line).

*_in this case it's fine, but generally you should __not__ do this_

Amend your commit with the corrections and force push. In the future, you'll want to create a feature branch for each of your pull requests, this is more for your own convenience than ours.

badshah400:
> Good catch, I really should go through all files to check for lingering non-Py3-compatible stuff.
> 
> Couple of things to fix.
> 
> **1.**
> The fix is wrong. `.open()` should be `.read()`.
> 

Sorry, that was careless. I should have just copied my patch from openSUSE where I did use `read()`.

> **2.**
> For the commit message we try to consistently use imperative mood - meaning that the commit message should be written like a command; "do x using y". Since this is a python 3 change, we would also use a category prefix to indicate the domain or reason for the change. In this case the message might be:
> 
> `Py3: use a string as the 1st argument to exec`
> 
> Since this is a one-liner that does not touch anything used by the ci, you can* add `[skip ci]` to the end of the message (only works if it is on the first line).
> 
> *_in this case it's fine, but generally you should **not** do this_
> 
> Amend your commit with the corrections and force push.

Thanks, done now.

PR #1065, opened 2020-03-18
Fix rounding in conversion to rgba8 while saving

gwojcik:
Each load, save cycle introduced error while converting internal data to rgba8.
Errors were noticeable after a couple of saves.
It caused loss of color details and banding artifacts.


gwojcik:
Examples of errors:
original image:
obraz
after 5 saves:
obraz
after 25 saves:
obraz

jplloyd:
Confirmed that this is a problem (except when in 1.x mode), and that the change fixes it. Thanks!

This will be backported to the 2.0.x branch to be included in a (not-too-distant) 2.0.1 release, as this is a fairly serious bug.

PR #1100, opened 2020-08-16
Line mode - move ends of cruve

gwojcik:
Changes in line mode, allow moving ends of curve by pressing shift + ctrl.


jplloyd:
The Hound is being annoying, but flake8 is happy with the changes, and so am I.

PR #1101, opened 2020-08-16
Fix hound config [skip travis] [skip appveyor]

jplloyd:
Duplication is not nice, so the current flake8 config in `setup.cfg` should prob. be removed.


jplloyd:
Some weirdness going on here, creating additional PR to verify.

Issue #899, opened 2018-02-22
AttributeError: 'EventGrabBroken' object has no attribute 'time' (when starting MyPaint)


kelvinwop:
#### Description of the problem

old version of mypaint threw error upon starting. Never got this error before, and could not reproduce. Logging it here just in case it is useful in the future.

#### Basic system details

MyPaint version: 1.3.0-alpha+git.aefccd44 (Python 2.7.14, GTK 3.22.21, GdkPixbuf 2.36.10, Cairo 1.15.6, GLib 2.54.0)
Operating system: win 8.1
Desktop environment: ??

#### Steps to reproduce

1. Start MyPaint
2. ☞☞

#### Backtraces or error messages

```python
Mypaint version: 1.3.0-alpha+git.aefccd44
System information: Windows-8.1-6.3.9600-SP0
Using: Python 2.7.14, GTK 3.22.21, GdkPixbuf 2.36.10, Cairo 1.15.6, GLib 2.54.0
Traceback (most recent call last):
  File "C:/Program Files/mypaint-git-w64/mingw64/lib/mypaint/gui/picker.py", line 364, _in_grab_grab_broken_cb(self=<gui.picker.ContextPickingGrabPresenter object>, widget=<drawwindow.DrawWindow object at 0x000000000d2a7b40 (MyPaintDrawWindow at 0x000000000d23cca0)>, event=<Gdk.EventGrabBroken object at 0x000000000f652050 (void at 0x0000000010bfad30)>)
            logger.debug("Grab broken, cleaning up.")
            self._ungrab_grabbed_devices(time=event.time)
            return False
  variables: {'self._ungrab_grabbed_devices': ('local', <bound method ContextPickingGrabPresenter._ungrab_grabbed_devices of <gui.picker.ContextPickingGrabPresenter object at 0x000000000d2af910>>), 'time': (None, [])}
AttributeError: 'EventGrabBroken' object has no attribute 'time'
```

achadwick:
Unreproducible, but the error trace may be of some use here.

kelvinwop:
Yeah, I couldn't reproduce it no matter what I tried. I posted it here anyways for future reference

PR #1104, opened 2020-08-26
Remove reference to non-existent Gdk.EventGrabBroken time attribute

calwas:
PyGTK Gdk.EventGrabBroken maps to the GdkEventGrabBroken struct which
does not have a time attribute. The invalid attribute was being passed
as an argument that had an appropriate default value, so the attribute
reference was removed, enabling the default value to be used instead.

Closes mypaint/mypaint#899


jplloyd:
Good catch, thanks!

(I rebased some translations onto your commit to avoid triggering the ci build twice.)

PR #1108, opened 2020-09-06
fix dimming of main window after dismissing recover backup window

gwojcik:
Problem was described by calsioro on community forum: https://community.mypaint.org/t/main-window-stays-dim-after-dismising-the-recover-backup-window/2573


jplloyd:
Looks good, thanks! Not sure why that particular dialog unset the transient_for window.

PR #1115, opened 2020-10-12
Add Russian translation

OlesyaGerasimenko:


jplloyd:
Thanks for the translations, but do you think there is a better translation along the lines of "Digital painting software for artists" rather than Raster graphics editor? I think most people would associate editing more with the tools available in GIMP or Photoshop.

OlesyaGerasimenko:
> Thanks for the translations, but do you think there is a better translation along the lines of "Digital painting software for artists" rather than Raster graphics editor? I think most people would associate editing more with the tools available in GIMP or Photoshop.

I suppose it could be "Программа цифровой живописи для художников компьютерной графики".

jplloyd:
I guess I should have looked closer at the original English text (and the French translation). Your translation seems to match that (whether the original is a good description is a different question).

PR #1138, opened 2021-02-18
Added the ability to filter keymaps.

nyabkun:
search

### DIFF
--- a/gui/accelmap.py
+++ b/gui/accelmap.py
@@ -13,11 +13,13 @@
 
 from __future__ import division, print_function
 import logging
+import re
 
 from lib.gibindings import Gtk
 from lib.gibindings import Gdk
 from lib.gibindings import Pango
 from lib.gettext import gettext as _
+from lib.gettext import C_
 
 import lib.xml
 from lib.pycompat import unicode
@@ -43,13 +45,14 @@ class AccelMapEditor (Gtk.Grid):
 
     __gtype_name__ = 'AccelMapEditor'
 
-    _COLUMN_TYPES = (str, str, str, str, str, str)
+    _COLUMN_TYPES = (str, str, str, str, str, str, str)
     _PATH_COLUMN = 0
     _ACCEL_LABEL_COLUMN = 1
     _ACTION_LABEL_COLUMN = 2
     _SEARCH_TEXT_COLUMN = 3
-    _ACCEL_LABEL_SORT_COLUMN = 4
-    _ACTION_LABEL_SORT_COLUMN = 5
+    _FILTER_TEXT_COLUMN = 4
+    _ACCEL_LABEL_SORT_COLUMN = 5
+    _ACTION_LABEL_SORT_COLUMN = 6
 
     _USE_NORMAL_DIALOG_KEYS = True
     _SHOW_ACCEL_PATH = True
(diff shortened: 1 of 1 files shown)


gui/accelmap.py
```diff
@@ -125,6 +136,23 @@ def __init__(self):
...
+    def _filter_list(self, model, iter, data):
+        if not self.filter_txt:
+            return True
+
+        words = self.filter_txt.lower().split()
+        accel_label = lib.xml.escape(model[iter][0]).lower()
+        key = lib.xml.escape(model[iter][1]).lower()
+
+        for word in words:
+            if (not word in accel_label) and (not word in key):
```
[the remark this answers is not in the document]
nyabkun: Sorry, it looks like I need to tweak the code a bit more because the behavior when trying to set key bindings is strange.

nyabkun:
Sorry, it looks like I need to tweak the code a bit more because the behavior when trying to set key bindings is strange.

gui/accelmap.py
```diff
@@ -70,20 +71,32 @@ def __init__(self):
...
 
+        self.set_row_spacing(5)
+
         store = Gtk.ListStore(*self._COLUMN_TYPES)
         self._store = store
         self._action_labels = {}
         self._accel_labels = {}
 
+        self._filter_entry = Gtk.Entry()
+        self._filter_entry.set_placeholder_text(_('Filter'))
```
jplloyd, line 82: Import and use the `C_` function instead of this one, and provide "keymap-editor" (or similar) as the first argument. 

The first argument is the "context", distinguishing multiple appearances of the same word, as that word might have different translations in different situations, for some languages.

gui/accelmap.py
```diff
@@ -291,14 +342,14 @@ def _accel_edited_cb(self, cell, path, newname):
...
-        it = store.get_iter(treepath)
-        accel_path = store.get_value(it, self._PATH_COLUMN)
+        it = self._filter.get_iter(treepath)
+        accel_path = self._filter.get_value(it, self._PATH_COLUMN)
         accel_label = self._accel_labels[accel_path]
         action_label = self._action_labels[accel_path]
 
         editable.set_sensitive(False)
         dialog = Gtk.Dialog()
+        dialog.set_modal(True)
```
jplloyd, line 352: Why is it necessary for this dialog to be modal?

gui/accelmap.py
```diff
@@ -279,10 +296,44 @@ def _get_accel_map_entries(cls):
...
+    def _filter_check(self, model, iter, search_key):
+        if search_key:
+            txt = search_key
+        else:
+            txt = self._filter_txt
+
+        if not txt:
+            return True
+
+        reg_sep = r'(?:\+(?!\w(?:$|\s))|\s)'
```
jplloyd, line 313: Add a short comment explaining the logic here. It usually helps when you come back to the code half a year later, especially if that code involves python regexes.

jplloyd (changes requested):
Nice job! This is a very nice quality of life feature that is sorely missing from the shortcut editor. Only a couple of small things that need to be looked at before merging.

gui/accelmap.py
nyabkun: When clicking multiple times, a number of dialogs were appearing at the same time, so I thought this would be more correct behavior, and fixed it.

nyabkun:
Thank you very much for your kind attention.
Actually, this was my first pull request to a public project.
I'll try to fix it now that you've told me what I need to fix.

gui/accelmap.py
jplloyd: Fair enough, it does make sense in this case.

jplloyd (approved):

PR #1139, opened 2021-02-18
Keyboard Input Bug #974

### LINKED ISSUE #974: Shortcuts don't work in non-latin keyboard layouts
The title says it all, MyPaint reads localized text input instead of actual physical keys.

How to reproduce: switch to Cyrillic, try any key combination
Versions and systems: probably all of them (I tested 1.2.1 and 2.0.0 under Win32 and 1.2.0 under Debian 9 / GNOME).

The fix is to remove the event.group parameter from translate_keyboard_state in gui/keyboard.py.

nyabkun:
Fixed a bug in the following image.

108369487-f9d5a880-723e-11eb-836f-40b4aeb09c88

### DIFF
--- a/gui/accelmap.py
+++ b/gui/accelmap.py
@@ -432,7 +432,9 @@ class AccelMapEditor (Gtk.Grid):
             keymap.translate_keyboard_state(
                 event.hardware_keycode,
                 event.state,
-                event.group,
+                # https://github.com/mypaint/mypaint/issues/974
+                # event.group
+                1
             ))
         keyval = Gdk.keyval_to_lower(keyval)
         mods = Gdk.ModifierType(
--- a/gui/keyboard.py
+++ b/gui/keyboard.py
@@ -109,7 +109,9 @@ class KeyboardManager:
         # Reason: ignoring CAPSLOCK and checking if SHIFT was pressed
         state = Gdk.ModifierType(event.state & ~Gdk.ModifierType.LOCK_MASK)
         res = keymap.translate_keyboard_state(event.hardware_keycode, state,
-                                              event.group)
+                                              # https://github.com/mypaint/mypaint/issues/974
+                                              # event.group)
+                                              1)
         if not res:
             # PyGTK returns None when gdk_keymap_translate_keyboard_state()
             # returns false.  Not sure if this is a bug or a feature - the only


nyabkun:
Fixed a bug in the following image.

108369487-f9d5a880-723e-11eb-836f-40b4aeb09c88

jplloyd:
Thanks! I will try out this, and your other smaller patch tomorrow, and probably merge them directly (to avoid triple rebuilds - our builds are unfortunately very slow).

nyabkun:
The bug has not been fixed.

There is a 50/50 chance of garbled characters or no garbled characters at startup.

The garbled characters are cured when I set the IME to English keyboard setting, but I don't know the cause. I'll try to find out more.

nyabkun:
I'm confused.

The output is unstable, even with the same input and in the same environment.

```python
from lib.gibindings import Gdk

keymap = Gdk.Keymap.get_default()

bound, keyval, effective_group, level, consumed_modifiers = (
    keymap.translate_keyboard_state(
        65,
        0,
        0
    ))

print(f'keyval : {keyval}')
```

eXbOZrh

## Environment
- Windows 10
- Japanese Keyboard
- MSys2 MinGW
- Multi Monitor

nyabkun:
The following seems to be the reason you are using hardware_keycode.

```python
# Instead of using event.keyval, we do it the lowlevel way.
# Reason: ignoring CAPSLOCK and checking if SHIFT was pressed
```

nyabkun:
In my case, it seems to work correctly if the `group` passed to `keymap.translate_keyboard_state` is fixed to `1`.

I am not sure if this will work correctly in other environments.

jplloyd:
From what I can tell, this doesn't really solve the problem that @ystr had. At least not when running on linux with a cyrillic keymap (I will try it on windows later today).

On the other hand I'm not sure whether this is a wontfix, something that should be optional (configurable), or if I'm just testing it incorrectly. The reason I say that is because having shortcuts be based on the localized keyboard output _appears_ to be the norm on all major applications I have tested it on. It is of course possible that my tests just don't match the use case, due to differing setups.

By the way, your other patches seem to work great! If I can bother you with one thing, that would be to squash the commits in each PR so that there is one commit in each, and write `[skip ci]` at the end of the first line when you edit the commit message.
The reason for this is to make the commit history more readable, easier to navigate, and faster to bisect (when hunting bugs).

While I could squash them myself, there are two reasons for why it is better if you do it:
1. Your squashed commit will still be signed by you
2. Github will automatically recognize that this PR has been merged, even if I merge it remotely.

nyabkun:
I'm sorry, I didn't know that if I rename a branch during a pull request, the pull request will be dereferenced.

jplloyd:
> I'm sorry, I didn't know that if I rename a branch during a pull request, the pull request will be dereferenced.

No worries.

nyabkun:
Thank you for all you've taught me!

I'm not used to pull requests, so it's very helpful to get some pointers.

I squashed each of them into a single commit.

jplloyd:
Scratch what I wrote previously. My mistake was only switching the keymap (with setxkbmap) when testing. When testing by installing a Russian locale, and using the input switcher in Ubuntu's default desktop environment, it appears that your modification indeed fixes the problem!

Issue #427, opened 2015-09-15
Add multi-touch support for zoom and scroll


nonameentername:
I'm looking at adding multi-touch zoom and scroll to mypaint similar to the design seen here: https://www.youtube.com/watch?v=iicnVez5U7M

I was looking at the mypaint code and could not find the location of pen input and zoom.  Could someone point out the classes that should be modified to add this functionality?

achadwick:
Hi werner - thanks for the interest!

Sure, multi-touch gesture support is something that I'd really like (and can test!) MyPaint is in a beta development stage right now, with strings frozen for translation, but a simple implementation of this would be brilliant. Now follows the headdump!

GTK itself is evolving support for gesture events, new since GTK 3.14 (`GtkGesture` and its friends). Our upcoming release is slated to support GTK 3.10 or later, but I am perfectly happy to accept tested, bug-free, and ultra-simple code that carefully supports both versions and which doesn't introduce new UI strings. Multi-version support is something that's easy with Python! (And to be honest, 3.12+ is just *better* for paint programs - so the requirement's semi-fluid already to my mind.)

Input events from tablets are captured in [`gui/document.py`](https://github.com/mypaint/mypaint/blob/d9dfea9d89a76b72dc7332383031655cd0ccb9b5/gui/document.py) and delegated to the top mode on the mode stack. Modes correspond to toolbar tools, mostly, although an individual tool's main mode could in theory push another mode onto the stack while the user does something fiddly. In practice, even the inking and frame tools just have stateful main modes.

The base classes for a lot of modes can be found in [`gui/mode.py`](https://github.com/mypaint/mypaint/blob/d9dfea9d89a76b72dc7332383031655cd0ccb9b5/gui/mode.py). I will admit that the class hierarchy is a bit horrible and convoluted in there: it's my doing, and I'm happy to help explain the thicket.

Beware the cooperative `__init__`s and event hander methods.
Other modules exist. Note the messy metaclass-based registration.

Scrolling at present is implemented as a mixin, `gui.mode.ScrollableModeMixin`, because scrollwheel events have their own GTK event type, and thus their own capture method. Touch events are similar - they have their own "

Question: does every tool need to be zoomable or rotatable with gestures? Perhaps they all do, in which case you could probably get away with *not* delegating the touch events from the controlling gui.document.Document (warning: badly named class! it's an MVC controller for the document, and I want to go MVP with it at some stage...).

[...]

PR #1178, opened 2022-03-06
Multi-touch gestures for zoom/scroll/rotate

brentyi:
Hi!

Not sure if this project is still being maintained, but here's a basic+noninvasive implementation of gesture-based zoom/scroll/rotate operations that I've been using. Currently only enabled for the Freehand mode.

#427

### DIFF
--- a/gui/freehand.py
+++ b/gui/freehand.py
@@ -35,6 +35,7 @@ logger = logging.getLogger(__name__)
 
 class FreehandMode (gui.mode.BrushworkModeMixin,
                     gui.mode.ScrollableModeMixin,
+                    gui.mode.GestureModeMixin,
                     gui.mode.InteractionMode):
     """Freehand drawing mode
 
--- a/gui/mode.py
+++ b/gui/mode.py
@@ -23,8 +23,6 @@ from lib.gibindings import GLib
 
 import lib.command
 from lib.brushsettings import settings_dict
-from lib.document import Document
-from lib.layer.data import SimplePaintingLayer
 from lib.observable import event
 from lib.pycompat import add_metaclass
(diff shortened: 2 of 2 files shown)


ChengduLittleA:
hummm interesting... will it break anything?

brentyi:
I haven't had any issues; the main caveat is that the current gesture implementation isn't compatible with modes that inherit from `DragMode`. There's probably a simple fix for that, but I don't know Gtk or the mypaint code well enough to say for sure. :shrug:

odysseywestra:
Okay Appveyor is rebuilding the windows builds so you should be able to test it on windows too. Had a dependency issue that needed to be resolved.

brentyi:
Thanks @odysseywestra!

Not sure if/when I'll have time to revisit this, but some notes/updates for posterity:
- I tested this in Windows, and the gesture itself works. But there's a minor issue where a click event (and thus a "dot" of paint) can be triggered at the end of the gesture.
- I'd like to get gestures working for the "drag" modes, but there seems to be a fundamental incompatibility with the pointer grabs. I can get things working perfectly by commenting out the `pointer_grab()` and `pointer_ungrab()` calls, but it's not clear to me what that might break. Pointer grabs have been deprecated for a long time anyways, so maybe if this code is updated with current practices getting gestures working robustly will be much easier.
- The current way my changes are set up, where the gesture events are attached/detached in the interaction mode enter/leave methods, doesn't make much sense when multiple interaction modes are stacked. But keeping things this way because pointer grabs from any of the `DragMode` subclasses break gesture events even after the pointer is ungrabbed; if the pointer grabs can be removed the gesture event handling can likely follow the same pattern as the scroll events.

PR #1183, opened 2022-04-06
setuptools fixes

jtojnar:
Fix build and some warnings.


odysseywestra:
I'll merge in the request since I was able to confirm it on my fork as well. I'll ignore the error in Appveyor since that is different issue that will need to be addressed.

PR #1184, opened 2022-06-05
Timestamp

### LINKED ISSUE #1172: Suggestion: Use date instead of numbers and letters as scrap filenames
When saving scraps, I find it more sensible that the filename will have a date instead of random numbers and letters.

What do you think?

teleportingtortoise:
Solution for feature request https://github.com/mypaint/mypaint/issues/1172

The optional setting shouldn't conflict with most user's existing settings, but if anyone sees a potential conflict I'm happy to add a true default that is like the original and lacks any timestamp parsing of any kind.


odysseywestra:
@AesaraB I would look into if you can go ahead and merge these in. 

@teleportingtortoise do you need to address anything else in this PR or are you comfortable with how it is now?

@mypaint/developers Need a peer review to see if any of you have any comments.

PR #1185, opened 2022-06-06
Scrap folder

### LINKED ISSUE #375: Scraps prefix setting: split into a path chooser button and a filename-only text entry
Currently the scraps prefix setting is a single string like 

    ~/MyPaint/scraps

that encodes the prefix for eventual scrap files like

    /home/username/MyPaint/scraps020_a.ora

The single string nature of this is fairly nasty UX, since the user has to think about paths and prefixes. It's also a pig to verify from the programming side when we need the folder to be present and correct, different OSes need different prefixes, _and_ the final part ("scrap") should be made localizable.

It would be better to split this one setting into two, a folder selectable with a button and a file prefix that the user can type in. The setting itself should be split too, so that the default for the former can be OS-specific, and so that the default for the latter can be localized.

teleportingtortoise:
An attempt to address #375


teleportingtortoise:
noting for myself that there needs to be some kind of handler so that people's old settings get reformatted properly

odysseywestra:
@AesaraB I would look into if you can go ahead and merge these in. 

@teleportingtortoise do you need to address anything else in this PR or are you comfortable with how it is now?

@mypaint/developers Need a peer review to see if any of you have any comments.

teleportingtortoise:
As far as I can tell I covered what was necessary and tried to stay within the project's code style. I'll get my toolchain back up in case any issues arise though.

PR #1193, opened 2022-11-09
python: fix for Python 3.11

sgn:
- python 3 always open in universal mode, U is default mode in 3.0, and removed in 3.11
- mypaint doesn't use ld?n?gettext, so bind_textdomain_codeset isn't needed, that function is deprecated in 3.8 and is no-ops in 3.10 and removed in 3.11


odysseywestra:
I at least need one other python dev to verify there's not anything more to this before merging.

odysseywestra:
@sgn I will merge since I'm noticing failed builds on my local box and Python traceback references what you worked on to prevent it. Just be ready for any issues tickets that may arise with this PR.

Issue #1209, opened 2023-12-22
Can not open old ora files


mitvitaminen:
#### Description of the problem
As stated I can not open old ora files, but filemanagers preview shows file just fine.
Pls Note that I can not open the file in gimp either, so I suppose it is some kind of wrong formatted save file. Can I fix this or are my pictures lost? Maybe open the file in text browser and edit it there somehow. Gimp says it is unsupported file format with newest gimp version in arch repo.

#### Basic system details
MyPaint version: 2.0.1+gitexport.478280f
Operating system: Arch Linux 6.6.7-arch1-1
Desktop environment: Gnome Desktop

#### Steps to reproduce
1. Start MyPaint
2. Open File
or
1. Open file from filemanager

#### Backtraces or error messages

```python
Mypaint version: 2.0.1+gitexport.478280f
System information: Linux-6.6.7-arch1-1-x86_64-with-glibc2.38
Using: Python 3.11.6, GTK 3.24.39, GdkPixbuf 2.42.10, Cairo 1.18.0, GLib 2.78.0
Traceback (most recent call last):
  File "/usr/lib/mypaint/gui/tileddrawwidget.py", line 855, _state_changed_cb(self=<tileddrawwidget.CanvasRenderer object at 0x7f4e...ileddrawwidget+CanvasRenderer at 0x55de2563fcf0)>, widget=<tileddrawwidget.CanvasRenderer object at 0x7f4e...ileddrawwidget+CanvasRenderer at 0x55de2563fcf0)>, oldstate=<enum GTK_STATE_INSENSITIVE of type Gtk.StateType>)
                self._insensitive_state_content = None
            self.update_cursor()
  variables: {'self.update_cursor': ('local', <bound method DrawCursorMixin.update_cursor of <tileddrawwidget.CanvasRenderer object at 0x7f4ec3bf3a80 (gui+tileddrawwidget+CanvasRenderer at 0x55de2563fcf0)>>)}
  File "/usr/lib/mypaint/gui/tileddrawwidget.py", line 590, update_cursor(self=<tileddrawwidget.CanvasRenderer object at 0x7f4e...ileddrawwidget+CanvasRenderer at 0x55de2563fcf0)>)
                # Often a red circle with a diagonal bar through it.
                c = Gdk.Cursor.new_for_display(
                    window.get_display(), Gdk.CursorType.CIRCLE)
[...]

PR #1235, opened 2024-02-03
Use standard cursor name

jtojnar:
Adwaita, the resident icon theme on GNOME removed the legacy X11 `circle` cursor: https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/commit/fb133a3989ee155c24201e30bab7dbc1c7c843fd

As a result, MyPaint would break when opening a file on GNOME systems.

Let’s use the equivalent `not-allowed` cursor from the XDG cursor-spec draft: https://www.freedesktop.org/wiki/Specifications/cursor-spec/

The circle cursor that was removed from Adwaita is in the fourth row, second column here: https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/blob/4d082e44e8361ca32360bb5b865be98a784ef4db/src/cursors/adwaita.svg
Here is the list of cursor constants supported by GDK 3 with images – it includes legacy X11 cursors: https://docs.gtk.org/gdk3/enum.CursorType.html
Here is the list of cursor names – only contains standard ones: https://docs.gtk.org/gdk3/ctor.Cursor.new_from_name.html

There was also simila  issue with the  `exchange` legacy X11 cursor when attempting to mask HCY wheel so Ifixed that as well.

And while at it, I switched to the preferred `Gdk.Cursor.new_from_name` API.

Fixes: https://github.com/mypaint/mypaint/issues/1209

### DIFF
--- a/gui/brushmanip.py
+++ b/gui/brushmanip.py
@@ -201,7 +201,7 @@ class BrushResizeMode(gui.mode.OneshotDragMode):
 
     @property
     def active_cursor(self):
-        return Gdk.Cursor.new(Gdk.CursorType.BLANK_CURSOR)
+        return Gdk.Cursor.new_from_name(Gdk.Display.get_default(), "none")
 
     def enter(self, doc, **kwds):
         super(BrushResizeMode, self).enter(doc, **kwds)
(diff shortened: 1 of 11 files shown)


AesaraB:
Hi, can you give me an .ora file that would break as per #1209 for me to test this on my side?

jtojnar:
It happens with any .ora file for me. Maybe try locking a layer.

AesaraB:
I can reproduce the bug by locking a layer. Looks like you've removed several calls to a depreciated method, thanks for that.

I can confirm this to be working on my side. Merging it into the codebase.

Issue #1216, opened 2024-01-17
Improve directory structure


AesaraB:
# Motivations
New contributors should need to know as little about the code base as possible before they can start hacking code, this means that the project should be structured in a way that makes it as easy as possible to form a mental model of the scope of each directory.

# Proposal
- [x] `build-aux` – files for building packages and development
	- appimage
	- flatpak
	- glade – Glade does not support GTK 4 so this can be removed
	- windows
- [ ] `data` – non-code files
	- backgrounds
	- desktop
	- palettes
	- pixmaps
	- [ ] `po` – translations
	    - [Weblate](https://hosted.weblate.org/projects/mypaint/) hooks to this directory. Changes in the source tree need to be reflected there.
	- svg
- [ ] `doc`
- [x] `src`
	- gui
	- lib
- [x] `tests`

AesaraB:
# Current issues
- Can't call /src/setup.py agonistically of working directory
  - I don't know where build artefacts should be placed (e.g. ``setup.py demo``)
- How do we separate program code from assets?
  - To be fair, the code exclusively resides in ``/src/lib/`` and ``/src/gui/``

jtojnar:
I think the current layout is pretty clean compared to e.g. libmypaint.

I propose the following change to use same layout as most other GTK apps (e.g. [Builder](https://gitlab.gnome.org/GNOME/gnome-builder)):

- `build-aux` – files for building packages and development
	- appimage
	- flatpak
	- glade – Glade does not support GTK 4 so this can be removed
	- windows
- `data` – non-code files
	- backgrounds
	- desktop
	- palettes
	- pixmaps
	- svg
- `doc`
- `src`
	- gui
	- lib
- `po` – translations
- `tests`

`setup.py` would remain in the root as the build system entry point.

AesaraB:
This is my first GTK project so I really appreciate your input -- this directory structure looks seriously good!

I would perhaps move ``po`` to ``data`` as most users won't use the directory frequently, but at the same time maintaining consistency with other projects would make it easier for contributors who are used to working with GTK acclimate to our codebase.

jtojnar:
Yeah, I would like moving `po` to `data` as well but decided to keep it in root on consistency grounds for now.

AesaraB:
I'm in the process (beginning, really) of vastly improving the developer documentation, so I think if we're both in agreement here we can make up for discrepancies in consistency with improved documentation.

[...]

PR #1242, opened 2024-02-24
Move build-related directories to build-aux

AesaraB:
Part of #1216, this PR cleans up the root directory a bit.


AesaraB:
@odysseywestra you're up

odysseywestra:
Okay, so the whole goal of this PR is to reduce the amount of .md docs in the base folder and move anything that packages or tests mypaint under a new folder?

The only thing I suggest doing is moving the `appveyor.yml` file to `/build-aux/windows/` too since I plan on replacing that with GitHub Actions build environment and it doesn't need to be hidden either. Doesn't have to be moved now, but I'll leave that up to you.

Other than that it looks solid and I'm ready to merge to the feature branch.

AesaraB:
> The only thing I suggest doing is moving the appveyor.yml file to /build-aux/windows/ too since I plan on replacing that with GitHub Actions build environment and it doesn't need to be hidden either.

I'll just delete appveyor.yml, since we won't merge [``feature-new-build``](https://github.com/mypaint/mypaint/pull/1244) until we also get the new CI pipeline up and running.

You'll have the master branch to refer to, and of course if future reference is needed we'll use the "version control" part of "distributed version control system" for that.

Issue #1223, opened 2024-01-18
Consolidated: Testing, Building, and Packaging for v2.1.0


AesaraB:
### Point of concern

The testing and building workflow is full of tech debt. If I'm going to make sweeping changes to MyPaint, I need these to work the best they can.

Additionally, there should be a explicit 'list' of things confirm as working/not-working prior to merging into a development branch, or merging into master.

#### Building:
- #1229
- #1216

#### Unit Testing:
- #1191
- #32 
- #461

#### Continuous Integration:
- #1238 
- #784

~~Realistically there will be two passes of this issue: One before #1216, and one after.~~ #1216 will be completed first.

AesaraB:
Succeeded by #1244

Issue #1191, opened 2022-08-31
Please migrate away from nose unittest framework


hosiet:
#### Description of the problem

Currently mypaint uses nose for unittests via `python3 setup.py nosetests` command. However, the original `nose` framework has ceased its development since 2015, and will be removed from Debian (and derivatives). See https://bugs.debian.org/1018417 for more information.

As a result, please consider migrating away from original nose onto other active projects, such as nose2, pytest, or python3-builtin unittests framework.

odysseywestra:
I'll probably adjust the yml files to use nose2 for testing so we can get builds going again.

odysseywestra:
https://docs.pytest.org/en/7.1.x/how-to/nose.html

PR #1245, opened 2024-02-24
Move to Meson Build System

AesaraB:
Completes #1223


AesaraB:
uh... @odysseywestra I literally just renamed the files and hound seems to be upset with me..

AesaraB:
# Dependencies
Meson can check for dependencies, this is what we've checked so far:
- [x] python
- [x] gtk3
- [x] pygobject
- [x] lcms
- [ ] gettext
- [ ] librsvg2
- [ ] numpy
- [ ] pycairo

AesaraB:
## [``build_ext``][stBuildExt]
- build C/C++ extensions (compile/link to build directory)
## [``build_py``][stBuildPy]
This is probably the simplest step in the build process. It copies files from two directories, ``gui/`` and ``lib/``

List of files copied by build_py: [build_py.txt](https://github.com/mypaint/mypaint/files/14392043/build_py.txt)

### ``gui/``
There are four "file types" in this directory. From what I gather, ``build_py`` just copies it all
- \*.glade
- \*.py
- \*.xml
- directories

### ``lib/``
There are four "files types" in this directory. ``build_py`` only copies all ``*.py`` files, including ones in subdirectores.
- \*.hpp
- \*.cpp
- \*.py
- directories

### The pattern?
There are two potential matching strategies for the pattern we see: additive, or subtractive.
1. Additive: \*.glade, \*.py, \*.xml are copied to the build directory
2. Subtractive: All files that aren't \*.hpp or \*.cpp are copied to the build directory

### Translating to Meson
You can't glob files in meson like you can in setuptools/distutils. Obviously this has implications for how we compile C/C++ files, but it also poses an issue for ``build_py``.

We can either:
1. [Individually specify all the ``lib/*.py`` files in meson.build](https://mesonbuild.com/Python-module.html#install_sources),
    - This may create an arduously long ``meson.build`` file.
2. Separate C/C++ files from python files in ``lib/``, or
3. Create a script to glob these files and modify ``meson.build`` for us.
    - I'm not sure how different this is in practice to option 1.

[stBuild]: https://github.com/pypa/distutils/blob/main/distutils/command/build.py#L18
[stCommand]: https://github.com/pypa/distutils/blob/main/distutils/cmd.py#L17
[stBuildExt]: https://github.com/pypa/distutils/blob/main/distutils/command/build_ext.py
[stBuildPy]: https://github.com/pypa/distutils/blob/main/distutils/command/build_py.py
[stBuildCLib]: https://github.com/pypa/distutils/blob/main/distutils/command/build_clib.py
[stBuildScripts]: https://github.com/pypa/distutils/blob/main/distutils/command/build_scripts.py

AesaraB:
##  ``BuildConfig(Command)``
- ``lib/config.py`` is a file created from a template.
- The template (which will get filled in with information we provide) is ``config.py.in``
- The substituted information comes in the form of a ``dict``
- ``instantiate_template()`` takes a template and substitution dictionary as arguments to create to create an output file (``lib/config.py``)

### Translating to Meson
Meson does out of source builds, meaning that ``lib/config.py`` will not exist in the the source directory, but rather the build directory.

AesaraB:
[Back to top](#issue-2152133865)

# [``build_ext``][stBuildExt]
The gist of this class is that it basically takes in a list of C++ files alongside a list of flags and uses swig to compile them to the build directory.

Output of build_ext: [build_ext.txt](https://github.com/mypaint/mypaint/files/14396357/build_ext.txt)

[stBuildExt]: https://github.com/pypa/distutils/blob/main/distutils/command/build_ext.py

jtojnar:
Yeah, I would expect something like https://github.com/linux-nvme/libnvme/blob/bff7dda7e2a0f883d0b89e23fed725c916de3e61/libnvme/meson.build#L29-L45 to work.

meson.build
```diff
@@ -0,0 +1,111 @@
...
+swig_args = [ '--std=c++11',
+  '-Wall',
+  '-Wno-sign-compare',
+  '-Wno-write-strings',
+  '-D_POSIX_C_SOURCE=200809L',
+  '-DNO_TESTS',  # FIXME: we're building against shared libmypaint now
+  '-g',  # always include symbols, for profiling
+  ]
+
+if system_os == "darwin"
```
eli-schwartz, line 46: Lexer error:

```
meson.build:47:16: ERROR: Double quotes are not supported. Use single quotes.
if system_os == "darwin"
                ^
```

meson.build
```diff
@@ -0,0 +1,111 @@
+project('MyPaint', 'cpp',
+  version : 'v2.1.0',
+  default_options : ['warning_level=3', 'cpp_std=c++14'])
+
+# Build machine info
+system_os = system()
```
eli-schwartz, line 6: No such function, you probably meant:
```suggestion
system_os = host_machine.system()
```

meson.build
```diff
@@ -0,0 +1,111 @@
...
+    'lib/fill/floodfill.cpp',
+    'lib/fill/gap_closing_fill.cpp',
+    'lib/fill/gap_detection.cpp',
+    'lib/fill/blur.cpp',
+    'lib/fill/morphology.cpp'
+  ],
+  output:  ['mypaintlib.py', 'lib/mypaintlib_wrap.cpp'],
+  command: swig_cmd,
+  install: true,
+  install_dir: [python3.get_install_dir(pure: false, subdir: 'libnvme'), false],
```
eli-schwartz, line 81:
```suggestion
  install_dir: [py.get_install_dir(pure: false, subdir: 'libnvme'), false],
```

You named it such, previously.

meson.build
```diff
@@ -0,0 +1,111 @@
...
+  )
+
+mypaintlib_clib = python3.extension_module(
```
eli-schwartz, line 84: Ditto.

```suggestion
mypaintlib_clib = py.extension_module(
```

meson.build
```diff
@@ -0,0 +1,111 @@
...
+  '_mypaintlib',
+  mypaintlibb_swig[1],
```
eli-schwartz, line 86:
```suggestion
  mypaintlib_swig[1],
```

meson.build
```diff
@@ -0,0 +1,111 @@
...
+  dependencies : [py3_dep],
```
eli-schwartz, line 87:
```suggestion
  dependencies : [python_dep],
```

meson.build
```diff
@@ -0,0 +1,111 @@
...
+else
+  swig_cmd = [swig, '-python', '-o', '@OUTPUT1@', '@INPUT0@']
+endif
+
+mypaintlib_swig = custom_target(
+  'mypaintlib.py',
+  input: ['lib/mypaintlib.i',
+    'lib/gdkpixbuf2numpy.cpp',
+    'lib/pixops.cpp',
+    'lib/fastpng.cpp',
```
eli-schwartz, line 68: I think this has to go in src/meson.build (or src/lib/meson.build), since all the paths are in src/lib/* and it will be painful as-is.

meson.build
```diff
@@ -0,0 +1,111 @@
...
+    'lib/brushsettings.cpp',
+    'lib/fill/fill_common.cpp',
+    'lib/fill/fill_constants.cpp',
+    'lib/fill/floodfill.cpp',
+    'lib/fill/gap_closing_fill.cpp',
+    'lib/fill/gap_detection.cpp',
+    'lib/fill/blur.cpp',
+    'lib/fill/morphology.cpp'
+  ],
+  output:  ['mypaintlib.py', 'lib/mypaintlib_wrap.cpp'],
```
eli-schwartz, line 78: Just output it to mypaintlib_wrapper.cpp, exact paths don't really matter for a .cpp file anyway.

AesaraB:
@eli-schwartz I appreciate your review, but the latest commit was half copied and pasted code that I was yet to fully translate. Nonetheless, you have made some useful suggestions, which I'll add in when I can sit down and look at them more closely.

eli-schwartz:
Ahh, gotcha. :)

(I have selfish reasons for having a small interest in seeing mypaint be compiled by meson, as it's currently a pain to handle diverging expectations of PEP 517 vs traditional setup.py usage and packagers are being strongly encouraged to only use PEP 517, which doesn't work here. So I figured any small help I could provide might make that transition happen faster. Please do feel free to ask me for help.)

AesaraB:
@eli-schwartz I'm tremendously grateful for your interest in improving MyPaint -- thank you for your offer of help.

Would you mind if I call you in to be a reviewer at the final stages of the draft?

eli-schwartz:
I'd be thrilled to. :)

AesaraB:
[Back to top](#issue-2152133865)

#  ``BuildTranslations(Command)``
Inherits distutils' [Command](https://github.com/pypa/distutils/blob/main/distutils/cmd.py#L17)

## Concept

Effectively a prerequisite to [BuildConfig(Command)](https://github.com/mypaint/mypaint/pull/1245#issuecomment-1962900712)

meson.build
AesaraB: Fixed

meson.build
AesaraB: Fixed

meson.build
AesaraB: Fixed

AesaraB:
Adding a new build dependency, 'python3-polib'. Need to document this soon

AesaraB:
[Back to top](#issue-2152133865)

# Installation

One of the big things that need to be addressed is installation, here I'd like to talk with package maintainers to see what they want involved.

For reference, here's the output of ``python3 setup.py install --help``

```
Options for 'Install' command:
  --prefix                             installation prefix
  --exec-prefix                        (Unix only) prefix for platform-
                                       specific files
  --home                               (Unix only) home directory to install
                                       under
  --install-base                       base installation directory (instead of
                                       --prefix or --home)
  --install-platbase                   base installation directory for
                                       platform-specific files (instead of --
                                       exec-prefix or --home)
  --root                               install everything relative to this
                                       alternate root directory
  --install-purelib                    installation directory for pure Python
                                       module distributions
  --install-platlib                    installation directory for non-pure
                                       module distributions
  --install-lib                        installation directory for all module
                                       distributions (overrides --install-
                                       purelib and --install-platlib)
  --install-headers                    installation directory for C/C++
                                       headers
  --install-scripts                    installation directory for Python
                                       scripts
  --install-data                       installation directory for data files
  --compile (-c)                       compile .py to .pyc [default]
  --no-compile                         don't compile .py files
  --optimize (-O)                      also compile with optimization: -O1 for
                                       "python -O", -O2 for "python -OO", and
                                       -O0 to disable [default: -O0]
  --force (-f)                         force installation (overwrite any
                                       existing files)
  --skip-build                         skip rebuilding everything (for
                                       testing/debugging)
  --record                             filename in which to record list of
                                       installed files
  --user                               install in user site-package
                                       '/home/defuser/.local/lib/python3.11/si
                                       te-packages'
  --old-and-unmanageable               Try not to use this!
  --single-version-externally-managed  used by system package builders to
                                       create 'flat' eggs
```

AesaraB:
[Back to top](#issue-2152133865)

# InstallScripts

This one doesn't do much other than filling out a couple variables, so I'm approaching it in the same way I would do [``BuildConfig(Command)``](#issuecomment-1962900712).

The one advantage of the distutils build is that just injects *everything* at the top of the file, so if adding/removing new configuration values the programmer only has to do it in one place instead of multiple. Right now MyPaint build only installs two binary files, so it should be manageable. Maybe this is something that should be documented?

## Translating to Meson
- [x] Adds the shebang
- [ ] Adds version info variables
- [x] Specifies operating directories for MyPaint

AesaraB:
@eli-schwartz hi, [this comment](https://github.com/mypaint/mypaint/pull/1245#issuecomment-1962922002)  has a code block prepended with ``setuptools gcc command``. It contains a bunch of compiler flags that setuptools uses by default, but meson does not. Should we worry about any of them or is it safe to skip them?

AesaraB:
I'm at a point where literally the only thing I'm not convinced of is the performance what with how we've got different compiler flags and all.

Let's add  #1191 as a prerequisite.

eli-schwartz:
> @eli-schwartz hi, [this comment](https://github.com/mypaint/mypaint/pull/1245#issuecomment-1962922002) has a code block prepended with `setuptools gcc command`. It contains a bunch of compiler flags that setuptools uses by default (default i.e. we didn't manually specify them for mypaint), but meson does not. Should we worry about any of them or is it safe to skip them?

Somehow failed to notice the ping, oops.

You do not need to worry about this. setuptools by default passes in the same $CFLAGS that were used to build CPython (including some appended by the CPython autoconf configure script), which is generally something you don't want at all.

eli-schwartz:
In particular,

> ```
> -DOPENSSL_LOAD_CONF
> ```

!!! this is not the openssl project (and you have no code linking to openssl)

---

> ```
> -DNDEBUG
> -fasynchronous-unwind-tables
> -funwind-tables
> ```

Affects debug information and backtrace generation.


---

> ```
> -fno-semantic-interposition
> ```

Can be useful, it basically says that any symbols exported by your DSO shouldn't assume that someone else is LD_PRELOAD'ing an alternative. Python added this flag for its own use cases to allow more efficient calling between functions internal to libpython.so, which it does a lot of.

---

> ```
> -fstack-clash-protection
> -fstack-protector-strong
> -fstack-clash-protection
> -U_FORTIFY_SOURCE
> -D_FORTIFY_SOURCE=3
> ```
These are linux distro hardening flags.

---

> ```
> -fwrapv
> ```

This says to assume your code is invalid and doesn't conform to the C programming language specification, and to generate slower machine code that tries to work around that coding bug. CPython used signed overflow on addition / subtraction / multiplication operations a couple decades ago and found it easier to simply apply this flag and forget.

---

> ```
> -Werror=return-type
> ```

No effect on build artifacts at all. As a -W flag, it is solely a linter.

thesamesam:
Could I gently ping this? On our side, folks are looking to remove any packages which aren't PEP517 compliant, and this PR solves that (by not using setuptools for something it is (no longer?) interested in doing).

AesaraB:
@thesamesam This PR is essentially complete, I just need to update the compiler flags with Eli's recommendations.

I won't merge this PR until I repair our unit testing code to see if there's a performance drop. This may not happen for a while because I'm focusing on uni atm.

PR #1250, opened 2024-03-12
Tolerate tool name being `None`

ColinKinloch:
Here's a relevant link: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/286

### DIFF
--- a/gui/brushmanager.py
+++ b/gui/brushmanager.py
@@ -92,7 +92,7 @@ def _device_name_uuid(device_name):
     """
     if not PY3:
         device_name = utf8(unicode(device_name))
-    return unicode(uuid.uuid5(_DEVICE_NAME_NAMESPACE, device_name))
+    return unicode(uuid.uuid5(_DEVICE_NAME_NAMESPACE, device_name or ''))
 
 
 def _quote_device_name(device_name):
--- a/gui/device.py
+++ b/gui/device.py
@@ -643,7 +643,8 @@ def device_is_eraser(device):
         return False
     if device.get_source() == Gdk.InputSource.ERASER:
         return True
-    if re.search(r'\<eraser\>', device.get_name(), re.I):
+    name = device.get_name()
+    if name is not None and re.search(r'\<eraser\>', name, re.I):
         return True
     return False
 


odysseywestra:
Would it be better to give it an "Unknown Device" kind of output instead of None? That way the user knows there's a device connected, just there not tool_name assigned to it? If that doesn't matter then I'm fine pulling it in as is. @AesaraB what are your thoughts?

jtojnar:
> Would it be better to give it an "Unknown Device" kind of output instead of None? That way the user knows there's a device connected, just there not tool_name assigned to it?

It does not look like the name is used for anything user facing here:

- `_device_name_uuid` is used in https://github.com/ColinKinloch/mypaint/blob/9518ffd59a09e3989b0dea9e418d0b2272b3855f/gui/brushmanager.py#L995 and https://github.com/ColinKinloch/mypaint/blob/9518ffd59a09e3989b0dea9e418d0b2272b3855f/gui/brushmanager.py#L1007 to associate brush with device. IMO we should not be using the name there since if multiple devices lack a name, they will be stepping on each other’s toes.
- The eraser hack comes from https://github.com/mypaint/mypaint/commit/314f7c6409d0494c212796dd56e2bcbe4b75177f. Ignoring it when name is missing is probably fine. (Ideally, we would remove it, hoping it is no longer necessary, but that is unrelated to this PR.)

jtojnar (commented):
Would be nice to include the link to the wayland stuff in the commit message so that the reasoning is accessible in the future (e.g. with `git blame`).

odysseywestra:
Then I'm guessing the right course of actions instead of none is to generate a consistent random device ID then?

jtojnar:
Well, random id would not be able to match anything because it would change between storing the brush and matching it against device. Maybe we could use the device id but that would not work when there are multiple devices with same id – I assume pencil and eraser would have the same id. And it is optional too.

odysseywestra:
Yeah that sounds like that should be a better way to go then just putting an exception for a empty device name.

So I guess in terms of this pull request we would have to we would need to close it, and then open up another issue regarding this for porting to gtk4.

PR #1255, opened 2024-08-03
Fix exporting brushes with mtime 0

jtojnar:
On NixOS, exporting a brush group by pressing “Export as Zipped Brushset” action in its properties fails with:

    ValueError: ZIP does not support timestamps before 1980

The `gui.brushmanager.BrushManager.export_group` doctest fails with the same error.

This is because Nix forces `mtime` of package files to 0 (1970-01-01) to ensure the build outputs are consistently reproducible.

Let’s disable strict timestamps to clamp the out-of-range timestamps to ZIP epoch. That argument was actually introduced in Python 3.8 because of reproducible builds: https://bugs.python.org/issue34097

This raises minimum version of Python to 3.8.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -81,7 +81,7 @@ installed before you can build it.
 - setuptools
 - pygobject
 - gtk3 (>= 3.12)
-- python (>= 2.7.4)
+- python (>= 3.8)
 - swig (>= 3)
 - numpy
 - librsvg2 (and its svg gdk-pixbuf loader)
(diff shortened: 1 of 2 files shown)


BUILDING.md
```diff
@@ -81,7 +81,7 @@ installed before you can build it.
 - setuptools
 - pygobject
 - gtk3 (>= 3.12)
-- python (>= 2.7.4)
+- python (>= 3.8)
```
jtojnar, line 84: This should be fine since even [supported LTS Debian (Bullseye)](https://endoflife.date/debian) ships at [least 3.9](https://wiki.debian.org/Python).

martinxyz (approved):

AesaraB (approved):
LGTM, merging

PR #1259, opened 2024-08-04
Fix some deprecations

jtojnar:


lib/strokemap.py
```diff
@@ -411,7 +411,7 @@ class _Tile:
 
     """
 
-    _ZDATA_ONES = zlib.compress(np.ones((N, N), 'uint8').tostring())
+    _ZDATA_ONES = zlib.compress(np.ones((N, N), 'uint8').tobytes())
```
martinxyz, line 414: Looks like `.tobytes()` is not even needed any more in this context. Not sure since when. But I guess it's safer to just migrate as you suggested.

```
In [6]: zlib.compress(np.ones((5,5)))
Out[6]: b'x\x9cc`\x00\x81\x0f\xf6\x0cC\x9c\x06\x00O\x14\x1d\x98'

In [7]: zlib.compress(np.ones((5,5)).tobytes())
Out[7]: b'x\x9cc`\x00\x81\x0f\xf6\x0cC\x9c\x06\x00O\x14\x1d\x98'
```

martinxyz (approved):

AesaraB (approved):
LGTM

AesaraB:
Force pushed to master, merging.

PR #1256, opened 2024-08-03
Remove Python 2 compatibility code

jtojnar:

PR #1268, opened 2024-09-23
Add type hints

jtojnar:
This will allow us to use [mypy](https://github.com/python/mypy) or [Pyright](https://github.com/microsoft/pyright) to statically check functions are actually called correctly.

This will require an extensive and careful review, it might be better to split it into several parts.

I used [pyment](https://github.com/dadadel/pyment) to convert the ReST docstrings to Google style, then [waterloo](https://github.com/anentropic/python-waterloo) to transmute docstrings to type hint comments, and finally [com2ann](https://github.com/ilevkivskyi/com2ann) to convert those to regular type hints.

Depends on https://github.com/mypaint/mypaint/pull/1256, https://github.com/dadadel/pyment/pull/138


QuLogic:
You'll probably want to fix the commit author on the `pymentize` commit.

PR #1281, opened 2025-01-31
sliderwidget.py: Use ParamSpec func over attr

AesaraB:
PyGObject doesn't cast the right type on ParamSpec instances whose attributes are referred to in a direct manner. Let's use ParamSpec methods instead.
https://lazka.github.io/pgi-docs/GObject-2.0/classes/ParamSpec.html

Also get rid of the hack that previously added GObject types (6a2f91a3)

@jtojnar ready for review

### DIFF
--- a/gui/sliderwidget.py
+++ b/gui/sliderwidget.py
@@ -33,21 +33,21 @@ class ScaleDelegator(type(Gtk.Bin)):
 
     def __init__(cls, name, bases, dict):
         # Existing properties from shared ancestry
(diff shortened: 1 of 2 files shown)


AesaraB:
Looks like our testing environment doesn't like my use of get_name((

AesaraB:
Tragically, "works on my system"

AesaraB:
Oh, looks like our test workflow uses an old version of Ubuntu

https://github.com/mypaint/mypaint/blob/master/.github/workflows/test.yaml#L10

PR #1282, opened 2025-01-31
main.py: Remove SIGINT workaround

### LINKED ISSUE #1278: Remove Ctrl + C SIGINT workaround
### Point of concern

Apparently at one point Ctrl + C didn't send the SIGINT signal to the process or something? Well, the [bug referred to in the code](https://bugzilla.gnome.org/show_bug.cgi?id=622084) is seemingly fixed now

https://github.com/mypaint/mypaint/blob/00483b9eed5240efcb8c807a40d68b9a823fe3da/gui/main.py#L228-L233

### Suggested improvements

Remove the code, test if Ctrl + C sends SIGINT

AesaraB:
The bug is fixed now
https://bugzilla.gnome.org/show_bug.cgi?id=622084

Closes #1278

### DIFF
--- a/gui/main.py
+++ b/gui/main.py
@@ -225,11 +225,6 @@ def main(
             func = app.filehandler.confirm_destructive_action
             gtkexcepthook.quit_confirmation_func = func
 
-        # temporary workaround for gtk3 Ctrl-C bug:
-        # https://bugzilla.gnome.org/show_bug.cgi?id=622084
-        import signal
-
-        signal.signal(signal.SIGINT, signal.SIG_DFL)
         Gtk.main()
 
     if options.trace:


AesaraB:
So yes, it works. But I'm not thrilled about the fact that we print a traceback after sending the SIGINT. It may cause some chaos with bug reporters who are unfamiliar with the terminal.

jtojnar:
Looking at other GTK Python apps, they either still do this:

- https://github.com/GNOME/gnome-music/blob/db9b99b530cf1a7bab7d494e4c547b81f5a0a709/gnome-music.in#L95
- https://gitlab.gnome.org/jpu/cambalache/-/blob/f15a53e61c6e9e3e43b224ddbbedbdd974f786f1/cambalache/app/cambalache.in#L40
- https://github.com/GNOME/showtime/blob/9312a74d0486f3e3fd597caab19c5a2ed05e1019/showtime/showtime.in#L17
- https://github.com/GNOME/gnome-tweaks/blob/6ea86229617dd6a96c04fcc62c83cf873133f2c0/gnome-tweaks#L103

or, interestingly, some call `GApplication.quit`:

- https://github.com/quodlibet/quodlibet/blob/02505e48752b609dd7ecca25df6aaba69587eab2/quodlibet/exfalso.py#L32
- https://github.com/GNOME/meld/blob/d564a81305bca0a340644bbf6af020f04fab418a/bin/meld#L450

or even do some more complicated logic:

- https://github.com/GNOME/orca/blob/6326a929795e7d8ab13637d571bf0de732ac0f76/src/orca/orca.py#L201
- https://github.com/GNOME/gnome-keysign/blob/f7ed151306caf0f085d37c31e98e5373ce850e65/keysign/app.py#L339

PR #1300, opened 2025-06-22
Replace tostring() with tobytes() in stroke.py

askmeaboutlo0m:
Because numpy removed the former function (in a minor release) with numpy/numpy#28254. The function was an alias, so there's no change in behavior, i.e. it always returned bytes.

See also 2a92b6baf452aba2cff3cc0a7782b301da3933d7, where the same function was already replaced in strokemap.py.


AesaraB (approved):
LGTM

PR #1301, opened 2025-07-09
Remove hotfix from 5be6d448

AesaraB:
Commit https://github.com/mypaint/mypaint/commit/5be6d4489cb409e0f0d79a6073f525d0866ea41e fixes a config migration issue.

This fix was first released in version 1.2.0, which came out in 2015 (which is itself 2 years after the commit).

Should we remove it?


AesaraB:
We could alternatively create a new module e.g. "migration.py" dedicated to sequentially migrating versions.

user preference files could contain version metadata, and we could make a lookup for updating configs as needed.

I just don't like that the way this fix is currently implemented is very "hacky" and out of place within the context of application.py

PR #1306, opened 2025-07-20
Fix AttributeError by replacing value_name with name

Osaidgit:
This PR fixes AttributeError on Arch Linux by replacing deprecated 'value_name' with 'name'. Verified the fix works.


jtojnar:
Supposedly, this is incomplete, as the same fix was already reverted: https://github.com/mypaint/mypaint/pull/1305

@AesaraB do you have details? Maybe we just need to bump minimum pygobject3 version in `configure.ac`? Though we will probably want to support both at least for a while.

PR #1314, opened 2026-01-13
Fix CI Testing and Updated APIs for pygobject 

odysseywestra:
This is a two part kind of pull request. This PR should fix the errors of MyPaint failing its test environment with the current build process. The current CI chain will change with Meson, but we still need a working main branch till then. This adds a missing dependency and updates some API calls for pygobject in Ubuntu 24.04 environment. I will need to add another GitHub actions CI for testing against the latest package version, but this is a start to get back on track.

This is also testing the viability of using LLMs to diagnose and help me figure out what was breaking in our current test structure. I will need @AesaraB and @jtojnar help in reviewing this to make sure I did not miss anything. We also need to have a conversation of how to address "Vibe Coding" PRs and what we will and won't accept. 

I'm in the campground of the usefulness of LLMs and using them as a tool, but not a crutch. I think for now accepting Vibe Coded PRs that address small bugs or updating current coding is fine. Also, I can see the usefulness of Documentation a plus as well. With new features, I will shy away from them since that increases the risks of more mistakes like garbage code happening and the PRs of those tend to get large. If we stick with small fixes here and there, I am okay with those. We will need to add more checks and balances to keep the process smooth. With this I was using Warp.dev with Claude as the LLM as acknowledged in the commit history.

With regards with the PR, I did put the research in to understand what was happening, but I'm just an Admin. I need a coder's eye to help me keep the LLMs from straying too far from its scope. For myself, I will focus on keep the CI tooling updated with the current setup until @AesaraB is confident with the future meson builds.

Edit: I updated the comment so I can clearly state my case with dealing with "Vibe Coding" PRs.


gui/sliderwidget.py
AesaraB: Commits 698c665bd3d49a289c9994ac175f709f08908ed2 and c940f0a5b8cc47d179f2fd39916e78dfe89566ee are literally untrue and in contradiction of:

1. 4e9f4ef08a6f82a456bb7cc5c69c352b1aef8f54
2. https://api.pygobject.gnome.org/GObject-2.0/class-ParamSpec.html#gi.repository.GObject.ParamSpec.get_default_value
3. https://api.pygobject.gnome.org/GObject-2.0/class-ParamSpec.html#gi.repository.GObject.ParamSpec.get_name

lib/glib.py
AesaraB: This makes no sense, it replaces what should be the name of a directory with a literal number

https://github.com/mypaint/mypaint/blob/7e4ec24506aa465c4eb176f1dcc9b70c1a45c1e2/lib/glib.py#L189-L195

AesaraB:
These changes are downright insanity, especially since they manage to screw up the most basics of basics. MR rejected.

gui/sliderwidget.py
AesaraB: > So what version of pygobject do we need to meet?

The changes in 4e9f4ef quite literally exist to add compatibility with the latest versions of PyGObject

PR #1316, opened 2026-01-13
Fix AttributeError on pygobject < 3.51.0

jtojnar:
This extracts changes from c6cbb04c0842d92d8ed97d5d8e00ddfb6e2d9068 into a helper function that supports both versions and wraps changes from 7e4ec24506aa465c4eb176f1dcc9b70c1a45c1e2 in try block, falling back to the old one.

Using explicit block so that it can be easily dropped in the future.

### DIFF
--- a/gui/device.py
+++ b/gui/device.py
@@ -266,6 +266,13 @@ class Monitor(object):
             return
         settings = self._device_settings.get(device)
         if not settings:
+
+            try:
+                source_kind = source.name
+            except AttributeError:
+                # Compatibility with PyGObject < 3.51
+                source_kind = source.value_name
+
             try:
                 vendor_id = device.get_vendor_id()
                 product_id = device.get_product_id()
@@ -276,7 +283,7 @@ class Monitor(object):
             logger.info(
                 "New device %r" " (%s, axes:%d, class=%s, vendor=%r, product=%r)",
                 device.get_name(),
-                source.name,
+                source_kind,
                 num_axes,
                 device.__class__.__name__,
                 vendor_id,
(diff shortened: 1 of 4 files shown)


odysseywestra:
@jtojnar could you tag which issues this should address as well please?

odysseywestra (approved):

jtojnar:
I am not aware of any issues fixed by this.

AesaraB (commented):
@jtojnar Thank you for this patch. I'm not sure what I was thinking with the changes I implemented, and I'll try to keep this in mind when implementing new fixes.

The code LGTM, and I can confirm the changes working on both a recent roll of Tumbleweed, alongside Ubuntu 22.04

Do you have any advice for how to determine the oldest version of libraries to support (e.g. should we maintain support for LTS branches of distros)? I think it may be a good idea to formalise this, as I haven't seen any documentation left on the topic by the previous maintainers.

jtojnar:
I would expect there to not be material differences within a stable release.

AesaraB:
Okay -- then that leaves one final question: When do we remove the helper code? I've seen plenty of instances in our codebase already for workarounds and version support fixes that are 5-10 years old. I don't think it's an idyllic position to be in, so it sounds like a good idea to give these kinds of fixes an expiry date (maybe in doctests?) from the getgo

jtojnar:
I guess we could add some structured comments and have a script that would report oudated ones based on minimal versions of dependencies listed in `meson.build`. And then bump the versions after each release.

But hopefully, these hacks won’t be that common, going forward, so perhaps just removing them as we notice them when refactoring code would be enough.

Issue #753, opened 2016-12-15
Cannot get a circle with stylus, when mouse is connected with MyPaint 1.3.0-alpha+git.2c07a27b


DigitalKO:
#### Description of the problem
Cannot get the center of any circle with stylus, when mouse is connected with MyPaint 1.3.0-alpha+git.2c07a27b, even if I restrict mouse to specific tasks or ignored it completely.

Indeed, the center of the circle is always according to the position of the mouse cursor. (even if the mouse cursor is outside of the canvas)

#### Basic system details

MyPaint version:  1.3.0-alpha+git.2c07a27b
Operating system: Debian SID
Desktop environment: GNOME 3.22

#### Steps to reproduce

1. Start MyPaint
2. Choose to draw circle with the icon, with the pen
3. If a mouse is connected, the middle of the circle is according to the mouse, not the pen being used.

![screenshot from 2016-12-15 20-49-26](https://cloud.githubusercontent.com/assets/24569830/21241645/da3b30d0-c310-11e6-85ac-4dcba73f42a3.png)

Here, the mouse cursor is not visible, because of the screenshot, but it is located at the center of the draw circle.

achadwick:
Do you mean the (Circles and Ellipses tool](https://github.com/mypaint/mypaint/wiki/v1.2-Ellipses-and-Circles)?

What about the other line drawing tools?

DigitalKO:
Desktop environment: GNOME 3.22 under Wayland
Yes, I mean this tool, and all drawing tools have the same issue.

DariaLapin:
Hello peeps!
I have the same issue with my tablet (Aiptek Slim Tablet 600u Premium II) but different releases while using 1.2.1-0dev1ppa1~xenial1 and having Ubuntu 16.04, Kernel 4.4.0-62 (64bit).

I cannot use/install the 32bit version either on 64bit Ubuntu because of dependency's but I have used the 32 bit version (1.2.0) on 32bit Ubuntu  which does work flawless.
Any help would be awesome!

Thank you for your work, time and efford!

artm:
same under gnome 3.28.1 under wayland, mypaint 1.2.0+gitexport.f62444e

PR #1323, opened 2026-05-11
Fix line/ellipse tools using mouse instead of tablet pos

sandrolort:
When using line/eclipse tool with a tablet, The starting position is picked from the mouse, not the tablet.

https://github.com/user-attachments/assets/9aad8edc-46c0-4499-91d8-1b65798f4ffe

Fixed by passing on the event (with its x,y cords) that invoked the tool and using those instead of local_mouse_state.

https://github.com/user-attachments/assets/c79878c5-7d20-4a98-8877-69149a46975a


sandrolort:
https://github.com/mypaint/mypaint/issues/753

Seems to be an old issue. 
I don't know if it was fixed & broken again or never fixed but, either way, should help.

askmeaboutlo0m:
In case a reviewer isn't familiar: on some Wayland compositors, every input device gets its own cursor. Not sure why, the only reason I managed to find is that one person wanted to be able to pause a video with their mouse and take notes with their stylus. Obviously this causes loads of breakage because applications and frameworks assume there is only "the cursor".

You can also get multiple cursors in X11 if you try hard enough to configure it that way, but since nobody does that there's might not be any handling for it in GTK.

Issue #642, opened 2016-04-06
Add Support for Auto Detecting Different Keyboard Layouts


zb13y:
Is it possible to use keyboard shortcuts without switching to english keyboard layout? Some programs have such ability, e.g. Krita. A bit annoying to switch keyboard layout everytime i start program.

achadwick:
Hi there - thanks for the feedback!

This issue tracker is not for direct user support really. We try to limit it so that each thread is a single fixable issue which can be closed by changing the MyPaint code. I won't close this report straight away though because sometimes we can help. Often the feedback is useful too, and sometimes support requests can turn into stuff programmers can actually do something about!

We may not be able to help you straight away here, but don't be discouraged if we can't. Issue trackers don't get that many views, sadly. You may find you get more useful feedback from the [MyPaint Community Forums](http://community.mypaint.org/), from Twitter ([@MyPaintApp](https://twitter.com/MyPaintApp), #mypaint), on IRC ([#mypaint on freenode](irc://freenode/mypaint)), or on the [MyPaint mailing list](https://mail.gna.org/listinfo/mypaint-discuss).

achadwick:
Let's try to help though... :smile:

I assume you're using a Cyrillic keymap on Windows 10 based on recent reports of yours?

Answer: yes, but you'll have to assign the non-Latin keycaps to the command you want in Edit→Preferences:

![image](https://cloud.githubusercontent.com/assets/61299/14323014/bcda4df8-fc16-11e5-9b60-58d8b9aca33c.png)

(I can't type Cyrillic, so I can't show you how.)

You can search the list of commands by typing, but you need to click in it first. (Yes, this is poor UI/UX. Sorry. :disappointed: We have plans to enhance it: #623 )

* Does this dialog change the displayed key when you press a Cyrillic letter which you can type on your keyboard?
* Are you able to use the resulting key in the main window immediately afterwards?

achadwick:
**TRIAGE NOTE:** looks like a user support request, currently. This is the wrong forum for direct user support, reporter has been redirected to better places to ask. However this is potential FAQ info, so we might as well answer centrally.

Ideally, let's confirm that everything's working as it should with Windows non-Latin keymaps, then close. If we can't manage that, this issue should be closed after one week if no further information is forthcoming that would allow it to be progressed to `bug` or `enhancement` status.

zb13y:
[...]

PR #1327, opened 2026-07-10
Keyboard: make shortcuts work across keyboard layouts

pozitron57:
## AI authorship and review status

This pull request was prepared by **Chat-GPT 5.6 Sol**.

**The code has not been reviewed by a human.** Runtime testing was performed **only on macOS**. Windows and Linux behavior is covered by simulated unit tests, but has not been tested on those operating systems.

## Purpose

This PR is intended to address #642 and the related reports #974 and #1141.

MyPaint currently translates shortcut key events using the hard-coded keyboard group `1`. That group does not have the same meaning on every GDK backend. In particular, on the macOS Quartz backend it represents the Option layer of the active input source, rather than a second keyboard layout. As a result, MyPaint can interpret a physical key as a different character; for example, `S` may be translated as `ß`.

Hard-coding a group also prevents shortcuts assigned in the active non-Latin layout from taking priority.

## Changes

- Add a shared helper for converting GDK key events to normalized GTK accelerators.
- Try the active keyboard group first, preserving user-defined shortcuts in the current layout.
- On non-macOS GDK backends, try the other groups reported for the hardware key as fallbacks.
- On macOS, where Quartz does not expose other input sources as GDK groups, add a last-resort ANSI-US physical-key fallback for standard printable key positions.
- Use the same normalization logic for shortcut dispatch and the shortcut editor.
- Remove the unconditional `group=1` translation.

The active-layout accelerator always has priority. A fallback is considered only when no action is assigned to the active-layout accelerator.

## Validation

- Manually confirmed shortcut behavior in a locally built MyPaint on macOS with U.S. and Russian-PC input sources.
- Built successfully from source on macOS 26.3.1, arm64, with Python 3.11.
- Added 12 unit tests covering active-layout priority, non-Latin and physical-key fallbacks, modifiers, shifted symbols, multiple GDK groups, failed translations, and duplicate removal.
- `pytest -q tests/test_keyboardutils.py`: 12 passed.
- Black, Flake8, `py_compile`, and `git diff --check`: passed.

The complete existing pytest suite could not be collected in the headless automation environment because importing the macOS GDK backend aborts when no GUI session is available.

## Testing limitations

- No runtime testing has been performed on Windows or Linux.
- The macOS physical fallback covers standard ANSI-US printable key positions. Platform-specific extra keys, such as JIS-only keys, are not included.

Addresses #642, #974, and #1141.


AesaraB:
Hi, thank you for the PR; MyPaint is not currently accepting AI-generated code.

pozitron57:
> Hi, thank you for the PR; MyPaint is not currently accepting AI-generated code.

OK, no problem! The main intention was to share the code which solves the issue that almost prevented me from using MyPaint.

odysseywestra:
@pozitron57 

Thanks for the PR, however @AesaraB is right, we will not accept pure LLM/AI-generated code(vibe coded) work. We don't mind you using an LLM to assist you but I believe purely relying on is not the right way to go about it. Also Linux is the primary platform for MyPaint so your build has to work on there first. It needs to work on the latest version of Debian for stability and Archlinux for Edge Cases.

It come down to ownership and understanding what you did and fixing it if issues pop up that are related to your PR. We are a community driven project now so everyone who contributes need to take part ownership of what they contributed within reason.

So you are welcome to submit more PRs again, but this time thoroughly test it and make sure it not only builds and works on linux, but also you as the contributor take ownership of your work.

Same also goes for having your LLM writing your PR too. 

All-in-All take ownership and pride into your work and don't rely on your LLM to code/write for you.

pozitron57:
@odysseywestra 
Thanks for the detailed comment, I agree completely with your point. Actually I didn't expect the code to be merged, I just wanted to share it as I told. Maybe PR is not the best way to do it — sorry for that.

PR #1331, opened 2026-09-13
Fix SWIG 4.5 compatibility

felixonmars:
Replace old PyString_FromString call with PyUnicode_FromString. SWIG 4.5 removed its Python 2 compatibility aliases, exposing the undefined C API name while generating bindings for Python 3.


jtojnar (approved):
Yuck that SWIG was [shimming it](https://github.com/swig/swig/commit/79f7a2b7cb7ddc256eba09c8770133148025171d).

LGTM, thanks.

